PR #11, opened 2019-09-10
Resource based commands

evanlouie:
- Added a custom command dispatcher. This dispatcher allow for arbitrarily nested Command objects; enabling resource based commands.
- Moved the `init` command under `project`; now it is called via `spk project init` (a la resource based commands)


andrebriggs (commented):
@evanlouie what else can a `project` action do? Is there any roadmap on this potential other commands? Just want to make sure we aren't getting too concrete too fast

andrebriggs:
@evanlouie why is there an "examples" directory?

bnookala (approved):

evanlouie:
@andrebriggs No roadmap really. This refactor came out some discussions had during the code walk through yesterday. It was brought up by multiple people that we should really support resource based commands (a la azcli); which is something we previously couldn't do in the code setup. I moved the `init` function primarily to give an example of how creating nested sub-commands looks in this setup.

jmspring (approved):
In general, LGTM.  The refactoring to support resource based commands limits us to basically - "resource" "command" - rather than variable depth.  I believe this is fine for our project at the moment, but should be revisited if more flexibility is needed.

andrebriggs:
> @andrebriggs No roadmap really. This refactor came out some discussions had during the code walk through yesterday. It was brought up by multiple people that we should really support resource based commands (a la azcli); which is something we previously couldn't do in the code setup. I moved the `init` function primarily to give an example of how creating nested sub-commands looks in this setup.

Agreed on the resource based commands. Thanks for clarifying the nested sub-command

andrebriggs (approved):

PR #13, opened 2019-09-15
`spk deployment`: Adding init and get commands

samiyaakhtar:
- consuming npm package that contains data access layer for `spk deployment <command>` (corresponding [PR](https://github.com/microsoft/spektate/pull/33) in spektate)
- added `init` and `get` commands as a start
- (still need to add unit tests for this area)

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "test-watch": "jest --watchAll"
   },
   "devDependencies": {
+    "@types/cli-table": "^0.3.0",
     "@types/jest": "^24.0.18",
     "@types/js-yaml": "^3.12.1",
     "@types/node": "^12.7.4",
@@ -45,13 +46,9 @@
     ]
   },
   "dependencies": {
-    "@types/cli-table": "^0.3.0",
     "cli-table": "^0.3.1",
     "commander": "^3.0.1",
-    "fs": "0.0.1-security",
     "js-yaml": "^3.13.1",
-    "open": "^6.4.0",
-    "os": "^0.1.1",
     "shelljs": "^0.8.3",
     "spektate": "^0.1.2",
     "uuid": "^3.3.3",
--- a/package.json
+++ b/package.json
@@ -45,9 +45,15 @@
     ]
   },
   "dependencies": {
+    "@types/cli-table": "^0.3.0",
+    "cli-table": "^0.3.1",
     "commander": "^3.0.1",
+    "fs": "0.0.1-security",
     "js-yaml": "^3.13.1",
+    "open": "^6.4.0",
+    "os": "^0.1.1",
     "shelljs": "^0.8.3",
+    "spektate": "^0.1.2",
     "uuid": "^3.3.3",
     "winston": "^3.2.1"
   }
--- a/src/commands/deployment/helper.ts
+++ b/src/commands/deployment/helper.ts
@@ -1,0 +1,221 @@
+import Table from "cli-table";
+import * as fs from "fs";
+import * as os from "os";
+import Deployment from "spektate/lib/Deployment";
+import AzureDevOpsPipeline from "spektate/lib/pipeline/AzureDevOpsPipeline";
+import IPipeline from "spektate/lib/pipeline/Pipeline";
+import { logger } from "../../logger";
+
+let hldPipeline: IPipeline;
+let clusterPipeline: IPipeline;
+let srcPipeline: IPipeline;
+const fileLocation = os.homedir() + "/.Spektate";
+export let config: { [id: string]: string } = {};
+export enum OUTPUT_FORMAT {
+  NORMAL = 0,
+  WIDE = 1,
+  JSON = 2
+}
+
+export class Helper {
+  public static initializePipelines() {
+    srcPipeline = new AzureDevOpsPipeline(
+      config.AZURE_ORG,
+      config.AZURE_PROJECT,
+      false,
+      config.AZURE_PIPELINE_ACCESS_TOKEN
+    );
+    hldPipeline = new AzureDevOpsPipeline(
+      config.AZURE_ORG,
+      config.AZURE_PROJECT,
+      true,
+      config.AZURE_PIPELINE_ACCESS_TOKEN
+    );
+    clusterPipeline = new AzureDevOpsPipeline(
+      config.AZURE_ORG,
+      config.AZURE_PROJECT,
+      false,
+      config.AZURE_PIPELINE_ACCESS_TOKEN
+    );
+  }
+
+  public static verifyAppConfiguration = (callback?: () => void) => {
+    if (
+      config.STORAGE_TABLE_NAME === "" ||
+      config.STORAGE_TABLE_NAME === undefined ||
+      config.STORAGE_PARTITION_KEY === "" ||
+      config.STORAGE_PARTITION_KEY === undefined ||
+      config.STORAGE_ACCOUNT_NAME === "" ||
+      config.STORAGE_ACCOUNT_NAME === undefined ||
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -45,9 +45,15 @@
     ]
   },
   "dependencies": {
+    "@types/cli-table": "^0.3.0",
+    "cli-table": "^0.3.1",
     "commander": "^3.0.1",
+    "fs": "0.0.1-security",
```
evanlouie: Don't need, included in node stdlib

package.json
```diff
@@ -45,9 +45,15 @@
...
     "js-yaml": "^3.13.1",
+    "open": "^6.4.0",
+    "os": "^0.1.1",
```
evanlouie: Don't need, included in node stdlib

src/commands/deployment/helper.ts
```diff
@@ -0,0 +1,221 @@
...
+      }
+    }
+  };
+
+  public static configureAppFromFile = (callback?: () => void) => {
+    fs.readFile(fileLocation, (error, data) => {
+      if (error) {
+        logger.error(error);
+      }
+      const array = data.toString().split("\n");
```
edaena: Do we want to ignore case also?

package.json
```diff
@@ -45,9 +45,15 @@
     ]
   },
   "dependencies": {
+    "@types/cli-table": "^0.3.0",
```
evanlouie: Let's keep all types in `"devDependencies"` 🙏

src/commands/deployment/helper.ts
```diff
@@ -0,0 +1,221 @@
...
+let srcPipeline: IPipeline;
+const fileLocation = os.homedir() + "/.Spektate";
+export let config: { [id: string]: string } = {};
+export enum OUTPUT_FORMAT {
+  NORMAL = 0,
+  WIDE = 1,
+  JSON = 2
+}
+
+export class Helper {
```
evanlouie, line 38: It would be good to have all exported/public classes/functions/methods to have some accompanying jsdoc

src/commands/deployment/helper.ts
```diff
@@ -0,0 +1,221 @@
...
+    );
+    clusterPipeline = new AzureDevOpsPipeline(
+      config.AZURE_ORG,
+      config.AZURE_PROJECT,
+      false,
+      config.AZURE_PIPELINE_ACCESS_TOKEN
+    );
+  }
+
+  public static verifyAppConfiguration = (callback?: () => void) => {
```
evanlouie, line 66: docs plz

src/commands/deployment/helper.ts
```diff
@@ -0,0 +1,221 @@
...
+      Helper.configureAppFromFile(callback);
+    } else {
+      Helper.initializePipelines();
+      if (callback) {
+        callback();
+      }
+    }
+  };
+
+  public static configureAppFromFile = (callback?: () => void) => {
```
evanlouie, line 97: docs plz

src/commands/deployment/helper.ts
```diff
@@ -0,0 +1,221 @@
...
+        config[key] = value;
+      });
+      Helper.initializePipelines();
+      if (callback) {
+        callback();
+      }
+    });
+  };
+
+  public static writeConfigToFile = (configMap: any) => {
```
evanlouie, line 119: docs plz

src/commands/deployment/helper.ts
```diff
@@ -0,0 +1,221 @@
...
+      data += "\n" + key + "=" + configMap[key];
+    });
+    fs.writeFile(fileLocation, data, (error: any) => {
+      if (error) {
+        logger.error(error);
+      }
+    });
+  };
+
+  public static getDeployments = (
```
evanlouie, line 134: docs plz

src/commands/deployment/helper.ts
```diff
@@ -0,0 +1,221 @@
...
+        if (outputFormat === OUTPUT_FORMAT.JSON) {
+          logger.info(JSON.stringify(deployments));
+        } else {
+          Helper.printDeployments(deployments, outputFormat);
+        }
+      }
+    );
+  };
+
+  public static printDeployments = (
```
evanlouie, line 168: docs plz

src/commands/deployment/helper.ts
```diff
@@ -0,0 +1,221 @@
...
+        table.push(row);
+      });
+
+      logger.info(table.toString());
+    } else {
+      logger.info("No deployments found for specified filters.");
+    }
+  };
+
+  public static getStatus = (status: string) => {
```
evanlouie, line 253: docs plz

src/commands/deployment/index.ts
```diff
@@ -0,0 +1,9 @@
+import { Command } from "../command";
+import { getCommandDecorator } from "./get";
+import { initCommandDecorator } from "./init";
+
+export const deploymentCommand = Command(
```
evanlouie, line 8: docs plz

src/commands/deployment/init.ts
```diff
@@ -0,0 +1,80 @@
+import commander from "commander";
+import { logger } from "../../logger";
+import { config, Helper } from "./helper";
+
+export const initCommandDecorator = (command: commander.Command): void => {
```
evanlouie, line 9: docs plz

evanlouie (changes requested):

src/commands/deployment/helper.ts
```diff
@@ -0,0 +1,221 @@
...
+      if (callback) {
+        callback();
+      }
+    }
+  };
+
+  public static configureAppFromFile = (callback?: () => void) => {
+    fs.readFile(fileLocation, (error, data) => {
+      if (error) {
+        logger.error(error);
```
evanlouie, line 100: Probably want to throw an exception here after the error so it doesn't continue

src/commands/deployment/helper.ts
```diff
@@ -0,0 +1,221 @@
...
+    } else {
+      Helper.initializePipelines();
+      if (callback) {
+        callback();
+      }
+    }
+  };
+
+  public static configureAppFromFile = (callback?: () => void) => {
+    fs.readFile(fileLocation, (error, data) => {
```
evanlouie: If we're expecting the file to contain text, we should use `fs.readFile(fileLocation, 'utf8', (error, data)...`. This will automatically make `data` a string

src/commands/deployment/helper.ts
```diff
@@ -0,0 +1,221 @@
...
+      if (error) {
+        logger.error(error);
+      }
+      const array = data.toString().split("\n");
+      array.forEach((row: string) => {
+        const key = row.split(/=(.+)/)[0];
```
evanlouie, line 105: Isn't this just splitting on `=`? If so, should we just use `.split("=")`?

src/commands/deployment/helper.ts
evanlouie: We also probably wanna do a `.split(/\r?\n/)` instead of just "\n". Windows will throw in `\r` carriages

src/commands/deployment/helper.ts
samiyaakhtar: I think it is.. sorry about that, sometimes I find the most difficult way to _split a string_

samiyaakhtar: Actually, I remember now why I did that, often the access keys for storage may contain `=`, so just trying to split by the first occurrence of `=` in every line

evanlouie (approved):
lgtm 👍

edaena (approved):

PR #20, opened 2019-09-24
CLI - Project - Add Service, create directory and add starter pipelin…

mtarng:
…es.yaml file

initial PR for https://github.com/microsoft/bedrock/issues/577

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
   },
   "dependencies": {
     "cli-table": "^0.3.1",
-    "@types/mkdirp": "^0.5.2",
     "commander": "^3.0.1",
     "js-yaml": "^3.13.1",
     "shelljs": "^0.8.3",
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
   },
   "dependencies": {
     "cli-table": "^0.3.1",
+    "@types/mkdirp": "^0.5.2",
     "commander": "^3.0.1",
     "js-yaml": "^3.13.1",
     "shelljs": "^0.8.3",
--- a/src/commands/project/add-service.ts
+++ b/src/commands/project/add-service.ts
@@ -2,14 +2,11 @@ import commander from "commander";
 import fs from "fs";
 import yaml from "js-yaml";
 import path from "path";
-import shell from "shelljs";
-import mkdirp from "mkdirp";
+import shelljs from "shelljs";
 import { promisify } from "util";
 import { logger } from "../../logger";
 import { generateAzurePipelinesYaml } from "../../lib/fileutils";
 import { IBedrockFile, IMaintainersFile } from "../../types";
-
-// npm install @types/mkdirp
 
 /**
  * Adds the init command to the commander command object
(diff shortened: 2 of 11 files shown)


src/commands/project/add-service.ts
```diff
@@ -0,0 +1,103 @@
+import commander from "commander";
+import fs from "fs";
+import yaml from "js-yaml";
+import path from "path";
+import shell from "shelljs";
+import mkdirp from "mkdirp";
```
evanlouie: should keep on using shelljs instead. you can do `shell.mkdir('-p', 'your-dir')`

src/commands/project/add-service.ts
```diff
@@ -0,0 +1,103 @@
...
+import { promisify } from "util";
+import { logger } from "../../logger";
+import { generateAzurePipelinesYaml } from "../../lib/fileutils";
+import { IBedrockFile, IMaintainersFile } from "../../types";
+
+// npm install @types/mkdirp
```
evanlouie: dont need

src/commands/project/add-service.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  logger.info(`Adding Service: ${serviceName}, to Project: ${rootProjectPath}`);
+  logger.info(
+    `MaintainerName: ${maintainerName}, MaintainerEmail: ${maintainerEmail}`
+  );
+
+  // Mkdir
+  const newServiceDir = path.join(rootProjectPath, serviceName);
+
+  mkdirp(`${newServiceDir}`, function(err) {
```
evanlouie: use `shell.mkdir('-p', newServiceDir)` instead

package.json
```diff
@@ -47,6 +47,7 @@
   },
   "dependencies": {
     "cli-table": "^0.3.1",
+    "@types/mkdirp": "^0.5.2",
```
evanlouie: remove

evanlouie: Actually not sure how this is working because we dont have `mkdirp` as a dependency

evanlouie (changes requested):

src/commands/project/add-service.ts
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Adds the init command to the commander command object
+ *
+ * @param command Commander command object to decorate
+ */
+export const addServiceCommandDecorator = (
+  command: commander.Command
+): void => {
+  command
+    .command("add-service <service-name>")
```
timfpark: For the `deployment` command, we separate nouns and verbs into two separate parts, eg.

`spk deployment get`

but here we are combining them into one eg.:

`spk add-service`

I would propose that we agree on one way of doing this and one common set of verbs (where possible) such that the cli feels like a coherent whole in the same way that the `kubectl` tool does despite having been developed by hundreds of engineers.

timfpark (changes requested):
Let's agree on a common grammar for commands across the different teams - otherwise looks great to me.

src/commands/project/add-service.ts
```diff
@@ -0,0 +1,93 @@
...
+      } catch (err) {
+        logger.error(
+          `Error occurred adding service ${serviceName} to project ${projectPath}`
+        );
+        logger.error(err);
+      }
+    });
+};
+
+export const addService = async (
```
evanlouie: Add some JSDoc please 🙏 
We should standardize on having all export/public functions/classes on having some accompanying docs

evanlouie (changes requested):

src/commands/project/add-service.ts
mtarng: This command was initially under `spk project add-service`, but after speaking with @evanlouie we're moving service creation/management under it's own command; now it's `spk service create` etc.

evanlouie (approved):

bnookala (approved):

PR #25, opened 2019-09-26
Added az cli to ci

evanlouie:


yradsmikham (approved):

NathanielRose (approved):
LGTM- QQ: Would we need to maintain versioning for AZ cli?

evanlouie:
@NathanielRose since the pipeline installs it every run, it should always be running against latest anyways. Unless some major API change happens, we _should_ be in the clear. Probably should add some version matching/parsing for `az --version` in spk though

PR #28, opened 2019-09-26
Refactoring helper.ts to use only exported methods

### LINKED ISSUE #608: Refactor Class to only only exported methods
****As** a:** SPK developer

**I want:** the `class` created [here](https://github.com/CatalystCode/spk/blob/master/src/commands/deployment/helper.ts#L38) to be refactored to follow the project convention of using only exported functions (ie; no classes).

**So that:** All code in the codebase follows a standard style

**Describe the solution you'd like:** Remove the `Helper` class and turn all public methods to `export`ed functions (eg; `export const initializePipelines = () => ...`).

**Describe alternatives you've considered:** N/A

**Additional context:** N/A

samiyaakhtar:
Following the `const export` rule
Closes https://github.com/microsoft/bedrock/issues/608


edaena (approved):

src/commands/deployment/get.test.ts
```diff
@@ -4,12 +4,12 @@ import {
   enableVerboseLogging,
   logger
 } from "../../logger";
-import { Helper, OUTPUT_FORMAT } from "./helper";
+import * as Get from "./get";
```
samiyaakhtar, line 7: Having to do `import * as <something> from ...` because `jest.spyOn` only takes in a name of a module, and not an exported const member

src/commands/deployment/get.ts
```diff
@@ -1,10 +1,29 @@
...
 import commander from "commander";
+import Deployment from "spektate/lib/Deployment";
 import { logger } from "../../logger";
-import {
-  getDeployments,
-  OUTPUT_FORMAT,
-  verifyAppConfiguration
-} from "./helper";
+import { verifyAppConfiguration } from "./init";
+import { clusterPipeline, config, hldPipeline, srcPipeline } from "./init";
```
sarath-p: I think line 5 and 6 can be merged. If you are using all exported methods from ./init, _import *_ is another option.

src/commands/deployment/get.ts
```diff
@@ -78,3 +97,138 @@ function processOutputFormat(outputFormat: string): OUTPUT_FORMAT {
 
   return OUTPUT_FORMAT.NORMAL;
 }
+
+/**
+ * Gets a list of deployments for the specified filters
+ */
```
sarath-p, line 115: please add docs/comments for all parameters. How about adding _async_ support for this and other methods in this file.

src/commands/deployment/get.ts
```diff
@@ -78,3 +97,138 @@ function processOutputFormat(outputFormat: string): OUTPUT_FORMAT {
...
+      printDeployments(deployments, outputFormat);
+    }
+    return deployments;
+  });
+};
+
+/**
+ * Prints deployments in a terminal table
+ */
+export const printDeployments = (
```
sarath-p, line 154: Similar to above to add comments for parameters.

sarath-p (changes requested):
Looks good. Few comments on docs and async.

evanlouie (approved):
yay for no classes 👍

sarath-p (approved):
Looks great!

PR #31, opened 2019-09-27
Adding spk infra command - Deploying Azure resources through bedrock

### LINKED ISSUE #589: As a developer, I want to implement infra validations prior to deploying resources
**As a:**
Developer

**I want:**
a CLI to perform validation against a local environment to ensure all prerequisites are installed and authentication (to Azure) is valid

**So that:**
an environment is prepared to deploy resources

The CLI command should do the following:
  - [x] Verify prerequisites are installed (az cli, terraform, git, and helm)
  - [x] Authenticate with Azure
  - [x] Check to see that environment variables are set

E.g. `spk infra init`

**Additional context:**

**Acceptance Criteria:**

spk tool (CLI) can be used to validate prerequisites, check for Bedrock repo clone and environment variables for service principal.

Yes

NathanielRose:
- Checks if initialized bedrock Environment has been configured (Closes https://github.com/microsoft/bedrock/issues/589)
- Executes a Terraform Init on a bedrock template that has been mapped to a provided bedrock source directory


src/commands/infra/version.ts
```diff
@@ -0,0 +1,35 @@
...
+export const versionCommandDecorator = (command: commander.Command): void => {
+  command
+    .command("version")
+    .alias("v")
+    .description(
+      "Initialize will verify that all infrastructure deployment prerequisites have been correctly installed."
+    )
+    .action(async opts => {
+      const { silent = false, output } = opts;
+      shell.exec(
```
yradsmikham: Should we maybe move away from using shelljs? 

Since this is just performing a `terraform version`, I don't see this being too big of a problem.

src/commands/infra/README.md
```diff
@@ -0,0 +1,11 @@
+## `spk infra`
```
yradsmikham, line 1: Let's update documentation on how to use `spk infra create`

src/commands/infra/version.ts
NathanielRose: Deleting command since it was used for initial testing

src/commands/infra/README.md
NathanielRose: Updated!

yradsmikham (approved):

src/commands/infra/create.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  // Remove once check for Bedrock source is integrated with `spk infra init`
+  if (!fs.existsSync(".bedrock")) {
+    const cloneBedrock = await exec("git", [
+      "clone",
+      "https://github.com/microsoft/bedrock.git",
+      ".bedrock"
+    ]);
+  }
+  // Increasing time for Terraform Init
+  jest.setTimeout(99999);
```
andrebriggs: @NathanielRose how long do we expect this test to last?

src/commands/infra/init.ts
```diff
@@ -0,0 +1,94 @@
...
+export const initCommand = (command: commander.Command): void => {
+  command
+    .command("init")
+    .alias("i")
+    .description(
+      "Initialize will verify that all infrastructure deployment prerequisites have been correctly installed."
+    )
+    .action(async opts => {
+      try {
+        if (await validatePrereqs(binaries)) {
```
andrebriggs, line 29: Instead of nesting can put this all in one conditional? Also we don't use the bool return value from `validateEnvVariables(envVar)`

src/commands/infra/init.ts
```diff
@@ -0,0 +1,94 @@
...
+      return false;
+    }
+  }
+  logger.info(
+    emoji.emojify("Installation of Prerequisites verified: :white_check_mark:")
+  );
+  return true;
+};
+
+export const validateAzure = async (): Promise<boolean> => {
```
andrebriggs, line 69: This method dependences on `validatePrereqs` being called first with `az` as one of the inputs. Because of this tight dependency can you see use combining these methods somehow?

src/commands/infra/init.ts
```diff
@@ -0,0 +1,94 @@
...
+    });
+};
+
+export const validatePrereqs = async (
+  executables: string[]
+): Promise<boolean> => {
+  // Validate executables in PATH
+  for (let i of executables) {
+    try {
+      await promisify(child_process.exec)("which " + i);
```
andrebriggs, line 52: How do we plan to enforce particular versions of these prereqs?

src/commands/infra/init.test.ts
```diff
@@ -0,0 +1,45 @@
+import child_process from "child_process";
+import {
+  disableVerboseLogging,
+  enableVerboseLogging,
+  logger
+} from "../../logger";
+import { validatePrereqs } from "./init";
```
andrebriggs: Instead of this duplication can ew do the following?
`import { validatePrereqs, validateAzure, validateEnvVariables } from "./init";`
I'm not sure if this a Typescript best practice or not

src/commands/infra/create.ts
```diff
@@ -0,0 +1,136 @@
...
+    // Identify which environment the user selected
+    const EnvironmentPath = path.join(bedrockPath, templateEnvironment);
+    if (fs.existsSync(EnvironmentPath)) {
+      logger.info(
+        `Initializing Bedrock Template Environment : ${templateEnvironment}`
+      );
+      logger.info(EnvironmentPath);
+      process.chdir(EnvironmentPath);
+      // Terraform init in environment directory
+      const init = await exec("terraform", ["init"]);
```
andrebriggs, line 127: Seems like we have a need to encapsulate `terraform`. I'm thinking we could have a `terraform` specific interface that essentially acts like a wrapper or shim for all terraform commands. I think this will help with testing a lot.

andrebriggs (changes requested):

src/commands/infra/init.ts
yradsmikham: We are using nested if statements to allow the CLI to terminate if it errors out. So we only allow the CLI to proceed to the next function if it passes the previous one. We still are relying on a returned boolean in order to unit test `validateEnvVariables()`

src/commands/infra/init.ts
yradsmikham: Enforcing specific versions of prereqs will come in the next iteration of building the `init` command. We plan on removing the hardcoded binaries in the array and possibly storing versions in a separate file that is specific to a version of the spk tool. This is still a discussion to be had regarding compatibility with specific Bedrock releases as well.

src/commands/infra/init.ts
yradsmikham: When you say "this method", did you mean the method `validateAzure`? If so, the answer is no, because we want to enforce a "fail fast" scenario. In a situation where Azure CLI is NOT installed, fail, before running any `az` command.

src/commands/infra/init.test.ts
yradsmikham: Yup! This can be simplified.

src/commands/infra/create.test.ts
NathanielRose: Terraform Init can take anywhere from 3 - 10 seconds locally. I allotted 100 seconds for the `spk infra create` test for future features and return the timeout back to 5 seconds on line [27](https://github.com/CatalystCode/spk/pull/31/files#diff-3b3e8075814d4d73acccfb456945a4f0R27).

src/commands/infra/create.ts
NathanielRose: Agreed, for the second iteration, a terraform wrapper would configure the backend state with a `terraform init` and allow for more customization with an environment template for `terraform apply`.

src/commands/infra/create.ts
```diff
@@ -0,0 +1,136 @@
...
+import { logger } from "../../logger";
+import { exec } from "../../lib/shell";
+import path from "path";
+import { promisify } from "util";
+
+/**
+ * Adds the init command to the commander command object
+ *
+ * @param command Commander command object to decorate
+ */
```
evanlouie, line 17: Move this comment block to right above `export const createCommandDecorator`. Otherwise IDE's wont pickup this jsdoc is for that function

src/commands/infra/create.ts
```diff
@@ -0,0 +1,136 @@
...
+
+// Global vars that attain if spk infra init has been executed and the path to a bedrock template directory
+let InitIsDone: boolean = true;
```
evanlouie: Do we need this var? I don't see it changed anywhere

src/commands/infra/create.ts
```diff
@@ -0,0 +1,136 @@
...
+    .option(
+      "--serviceprincipalid <sp-id>",
+      "Service Principal ID for Azure Subscription"
+    )
+    .option(
+      "--serviceprincipalsecret <sp-secret> ",
+      "Service Principal Secret for Azure Subscription"
+    )
+    .action(async opts => {
+      console.log(opts.environment);
```
evanlouie: Remove or comment out please

src/commands/infra/create.ts
```diff
@@ -0,0 +1,136 @@
...
+        //await templateConfig();
+        //await templateDeploy();
+      } catch (err) {
+        logger.error("Error occurred while initializing");
+        logger.error(err);
+      }
+    });
+};
+
+export const validateInit = async (bedrockPath: string): Promise<boolean> => {
```
evanlouie, line 80: We should give some JSDoc for all `export'd functions

package.json
```diff
@@ -48,8 +48,10 @@
   "dependencies": {
     "azure-devops-node-api": "^9.0.1",
     "cli-table": "^0.3.1",
+    "@types/node-emoji": "^1.8.1",
```
evanlouie: We want to keep all `@types` in `devDependencies` instead of `dependencies`

Run:

```sh
$ yarn remove @types/node-emoji
$ yarn add -D @types/node-emoji
```

src/commands/infra/create.ts
```diff
@@ -0,0 +1,136 @@
...
+  try {
+    // TODO: Use this function to check the state of spk infra init and attain bedrock source location
+    if (fs.existsSync(bedrockPath)) {
+      if (InitIsDone) {
+        logger.info(
+          "`spk infra init` has been successfully executed, you may now proceed to deploy Bedrock environments."
+        );
+        return true;
+        logger.info(bedrockPath);
+        process.chdir(bedrockPath);
```
evanlouie: Would it be possible to change the working dir back to the original once the command finishes? Having a different working dir might break some commands down the line if we start composing them together

src/commands/infra/create.ts
```diff
@@ -0,0 +1,136 @@
...
+): Promise<string> => {
+  try {
+    // Identify which environment the user selected
+    const EnvironmentPath = path.join(bedrockPath, templateEnvironment);
+    if (fs.existsSync(EnvironmentPath)) {
+      logger.info(
+        `Initializing Bedrock Template Environment : ${templateEnvironment}`
+      );
+      logger.info(EnvironmentPath);
+      process.chdir(EnvironmentPath);
```
evanlouie: Any way to change back to original once command is completed?

src/commands/infra/init.ts
evanlouie, line 69: JSDoc plz 🙏

src/commands/infra/init.ts
```diff
@@ -0,0 +1,94 @@
...
+
+let binaries: string[] = ["terraform", "git", "az", "helm"];
+let envVar: string[] = [
+  "ARM_SUBSCRIPTION_ID",
+  "ARM_CLIENT_ID",
+  "ARM_CLIENT_SECRET",
+  "ARM_TENANT_ID"
+];
+
+export const initCommand = (command: commander.Command): void => {
```
evanlouie, line 20: Can you move the JSdoc to line 22 (right above the function definition)?

src/commands/infra/init.ts
```diff
@@ -0,0 +1,94 @@
...
+    await promisify(child_process.exec)("az account show -o none");
+  } catch (err) {
+    logger.error(emoji.emojify(":no_entry_sign: " + err));
+    return false;
+  }
+  logger.info(emoji.emojify("Azure account verified: :white_check_mark:"));
+  return true;
+};
+
+export const validateEnvVariables = async (
```
evanlouie, line 86: JSDoc plz 🙏

evanlouie (changes requested):

src/commands/infra/create.ts
NathanielRose: Currently being used as a holder until we have integration with checking global config file for bedrock prereqs.

src/commands/infra/create.test.ts
andrebriggs: Let's reduce the timeout and if we have issues we can increase it.

src/commands/infra/init.ts
andrebriggs: I understand. If we are modifying our prod code to accommodate tests in that manner it could indicate we can test in another way. Just a thought.

src/commands/infra/create.ts
NathanielRose: reverting back to original dir

src/commands/infra/create.test.ts
NathanielRose: Reduced to 10 seconds

src/commands/infra/README.md
```diff
@@ -0,0 +1,24 @@
...
+- Check for environment variables (e.g. ARM_SUBSCRIPTION_ID, ARM_CLIENT_ID,
+  ARM_CLIENT_SECRET, ARM_TENANT_ID)
+
+#### `spk infra create <environment>`
+
+Validates that a working bedrock source is available. The `spk infra create`
+will do the following:
+
+- Check if a working Bedrock source for infrastructure deployment is available
+- Checks if `spk infra init` has been ran (Work in Progress)
```
timfpark: ran -> run (grammar nitpick)

src/commands/infra/create.ts
```diff
@@ -0,0 +1,137 @@
+import commander from "commander";
+import fs, { chmod } from "fs";
+import { logger } from "../../logger";
+import { exec } from "../../lib/shell";
+import path from "path";
+import { promisify } from "util";
+
+// Global vars that attain if spk infra init has been executed and the path to a bedrock template directory
+let InitIsDone: boolean = true; //Holder until spk infra init integration
```
timfpark: Consider deciding on one code capitalization style (`InitIsDone` capitalized but `bedrockDir` and `rootDir` are camelcased)

timfpark (dismissed):

src/commands/infra/init.ts
yradsmikham: That's fair, and will definitely keep this mind. I have created a bool return with the intention that the `init` command will extend with additional functions.

andrebriggs:
@NathanielRose this might be a nitpick but I noticed that this PR increases overall test times from 10 seconds to 33 seconds. I'm looking at the tests output to try to figure out why

evanlouie:
> @NathanielRose this might be a nitpick but I noticed that this PR increases overall test times from 10 seconds to 33 seconds. I'm looking at the tests output to try to figure out why

@andrebriggs Thats probably from the cloning of the bedrock repo (its pretty big) and calling terraform init (which i've seen take 10-15 seconds on some slow machines)

src/commands/infra/create.ts
```diff
@@ -0,0 +1,137 @@
+import commander from "commander";
+import fs, { chmod } from "fs";
+import { logger } from "../../logger";
+import { exec } from "../../lib/shell";
+import path from "path";
+import { promisify } from "util";
+
+// Global vars that attain if spk infra init has been executed and the path to a bedrock template directory
+let initValidated: boolean = true; //Holder until spk infra init integration
```
evanlouie: Should `const` instead of `let` when the variable is not going to be redefined; The TSLint extension for VSCode should highlight this

src/commands/infra/create.ts
```diff
@@ -0,0 +1,137 @@
...
+let bedrockDir: string = `${process.cwd()}/.bedrock/cluster/environments`;
```
evanlouie: `const` instead of `let`

src/commands/infra/create.ts
```diff
@@ -0,0 +1,137 @@
...
+let rootDir: string = `${process.cwd()}`;
```
evanlouie: `const` instead of `let`

src/commands/infra/init.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+afterAll(() => {
+  disableVerboseLogging();
+  jest.setTimeout(5000);
+});
+
+describe("Validating executable prerequisites", () => {
+  test("Validate that array of executables do not exists in PATH", async () => {
+    // Iterate through an array of non-existent binaries to create a force fail. If fails, then test pass
+    let fakeBinaries: string[] = ["ydawgie"];
```
evanlouie: `const` instead of `let`

src/commands/infra/init.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    child_process.exec("az logout");
+    const value = await validateAzure();
+    expect(value).toBe(false);
+  });
+});
+
+describe("Validating environment variables", () => {
+  test("Test whether environment variables are set and not null", async () => {
+    // Set environment variables to null, and create a force fail scenario
+    let variables: string[] = ["ydawgie"];
```
evanlouie: `const` instead of `let`

src/commands/infra/init.ts
```diff
@@ -0,0 +1,104 @@
+import child_process from "child_process";
+import commander from "commander";
+import emoji from "node-emoji";
+import { promisify } from "util";
+import { logger } from "../../logger";
+
+let binaries: string[] = ["terraform", "git", "az", "helm"];
```
evanlouie: `const` instead of `let`

src/commands/infra/init.ts
```diff
@@ -0,0 +1,104 @@
...
+let envVar: string[] = [
```
evanlouie: `const` instead of `let`

src/commands/infra/init.ts
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Validates that prerequisites are installed
+ *
+ * @param executables Array of exectuables to check for in PATH
+ */
+export const validatePrereqs = async (
+  executables: string[]
+): Promise<boolean> => {
+  // Validate executables in PATH
+  for (let i of executables) {
```
evanlouie: `const` instead of `let`

src/commands/infra/init.ts
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Validates that environment variables are set and not null
+ *
+ * @param variables Array of environment vairables to check for
+ */
+export const validateEnvVariables = async (
+  variables: string[]
+): Promise<boolean> => {
+  // Validate environment variables
+  for (let i of variables) {
```
evanlouie: `const` instead of `let`

evanlouie (changes requested):

andrebriggs:
> @andrebriggs Thats probably from the cloning of the bedrock repo (its pretty big) and calling terraform init (which i've seen take 10-15 seconds on some slow machines)

@NathanielRose have we explored better ways to do this? Do we really want to test `terraform` here? My only concerns is if this if going to become a pattern of testing `terraform` calls

evanlouie (approved):

NathanielRose:
> > @andrebriggs Thats probably from the cloning of the bedrock repo (its pretty big) and calling terraform init (which i've seen take 10-15 seconds on some slow machines)
> 
> @NathanielRose have we explored better ways to do this? Do we really want to test `terraform` here? My only concerns is if this if going to become a pattern of testing `terraform` calls

There may be a better way to test this. when we introduce terragrunt for terraform state mapping. However, we do need a `terraform init` test for future support for providing an environment template through a git repo. `azure-simple` is our smallest bedrock repo that im validating otherwise I can simply map the provided environment template for `validateInit` to a bedrock folder and end the test.

I'm still thinking of a clean way to abstract that test out atm unless we are willing to take the hit and not have `templateInit` be tested?

andrebriggs:
> > > @andrebriggs Thats probably from the cloning of the bedrock repo (its pretty big) and calling terraform init (which i've seen take 10-15 seconds on some slow machines)
> > 
> > 
> > @NathanielRose have we explored better ways to do this? Do we really want to test `terraform` here? My only concerns is if this if going to become a pattern of testing `terraform` calls
> 
> There may be a better way to test this. when we introduce terragrunt for terraform state mapping. However, we do need a `terraform init` test for future support for providing an environment template through a git repo. `azure-simple` is our smallest bedrock repo that im validating otherwise I can simply map the provided environment template for `validateInit` to a bedrock folder and end the test.
> 
> I'm still thinking of a clean way to abstract that test out atm unless we are willing to take the hit and not have `templateInit` be tested?

OK. I think we really need to watch the complexity here going forward.

andrebriggs (approved):

PR #32, opened 2019-09-27
#605 Enhance List of Deployments with --watch option

edaena:
Refer to [microsoft/bedrock #605](https://github.com/microsoft/bedrock/issues/605)

### DIFF
--- a/src/commands/deployment/get.ts
+++ b/src/commands/deployment/get.ts
@@ -30,6 +30,8 @@ export enum OUTPUT_FORMAT {
   JSON = 2
 }
 
+const timeInterval = 5000;
+
 /**
  * Adds the get command to the commander command object
  * @param command Commander command object to decorate
@@ -178,8 +180,6 @@ export const watchGetDeployments = (
   service?: string,
   deploymentId?: string
 ): void => {
-  const timeInterval = 5000;
-
   setInterval(() => {
     getDeployments(
       outputFormat,
--- a/src/commands/deployment/get.ts
+++ b/src/commands/deployment/get.ts
@@ -30,8 +30,6 @@ export enum OUTPUT_FORMAT {
   JSON = 2
 }
 
-const timeInterval = 5000;
-
 /**
  * Adds the get command to the commander command object
  * @param command Commander command object to decorate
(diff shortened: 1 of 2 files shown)


src/commands/deployment/get.ts
```diff
@@ -146,6 +159,40 @@ export const getDeployments = (
...
+export const watchGetDeployments = (
+  outputFormat: OUTPUT_FORMAT,
+  environment?: string,
+  imageTag?: string,
+  p1Id?: string,
+  commitId?: string,
+  service?: string,
+  deploymentId?: string
+): void => {
+  const timeInterval = 5000;
```
samiyaakhtar, line 181: nit: Can we add this `timeInterval` to the top of the file so it's easy to find and edit?

samiyaakhtar (approved):

src/commands/deployment/get.ts
```diff
@@ -30,6 +30,8 @@ export enum OUTPUT_FORMAT {
   JSON = 2
 }
 
+const timeInterval = 5000;
```
evanlouie: can we move this into `watchGetDeployments`'s body or add it as an option for the function? don't like having floating top level variables unless multiple things need to use it.

evanlouie (changes requested):

src/commands/deployment/get.ts
evanlouie: If @samiyaakhtar wants to keep the variable on top level scope, can we add some doc around it to say what the variable does?

samiyaakhtar: I'm fine with keeping it in the body of the function if you don't prefer having floating top level const variables, i thought we wanted constants that change functionality to be more visible

edaena: I'll revert it for now. I think we can later decide on how to structure the code. I was talking to Samiya about how it would also be nice to have constant strings like error messages etc. in their own file. It helps with consistency and reusing them etc. It doesn't seem like a priority for now though. It's just something to think about.

samiyaakhtar: Sounds good, sorry about the confusion here @edaena

evanlouie (approved):

PR #41, opened 2019-10-02
refactor: add helper class for validating inputs

### LINKED ISSUE #634: [SPK] create helper methods for validating cli input
**As a:**
Developer

**I want:**
to be able to call helper that can validate input to check for empty/null/expression validity.  

**So that:**

helper methods 

**Additional context:**
adopted these methods from an existing project https://github.com/serverless/serverless-azure-functions/blob/master/src/shared/guard.ts

mydiemho:
fixes https://github.com/microsoft/bedrock/issues/634

### DIFF
--- a/src/commands/service/create.ts
+++ b/src/commands/service/create.ts
@@ -7,6 +7,7 @@ import {
   addNewServiceToMaintainersFile,
   generateAzurePipelinesYaml
 } from "../../lib/fileutils";
+import { guardNotEmpty, guardTypeOf } from "../../lib/guard";
 import { IUser } from "../../types";
 
 /**
@@ -41,26 +42,12 @@ export const createCommandDecorator = (command: commander.Command): void => {
       const projectPath = process.cwd();
       try {
         // Type check all parsed command line args here.
-        if (typeof serviceName !== "string") {
-          throw new Error(
-            `serviceName must be of type 'string', ${typeof serviceName} given.`
-          );
-        }
-        if (typeof packagesDir !== "string") {
-          throw new Error(
-            `packagesDir must be of type 'string', ${typeof packagesDir} given.`
-          );
-        }
-        if (typeof maintainerName !== "string") {
-          throw new Error(
-            `maintainerName must be of type 'string', ${typeof maintainerName} given.`
-          );
-        }
-        if (typeof maintainerEmail !== "string") {
-          throw new Error(
-            `maintainerEmail must be of type 'string', ${typeof maintainerEmail} given.`
-          );
-        }
+        const stringType = "string";
+        guardTypeOf(serviceName, stringType);
+        guardTypeOf(packagesDir, stringType);
+        guardTypeOf(maintainerName, stringType);
+        guardTypeOf(maintainerEmail, stringType);
+
         await createService(projectPath, serviceName, packagesDir, {
           maintainerEmail,
           maintainerName
--- a/src/lib/guard.ts
+++ b/src/lib/guard.ts
@@ -1,0 +1,42 @@
+/**
+ * Validates the string express is not null or empty, otherwise throws an exception
+ * @param value - The value to validate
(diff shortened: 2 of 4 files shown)


src/lib/guard.ts
```diff
@@ -0,0 +1,42 @@
...
+  message?: string
+) => {
+  if (!!value === false) {
+    message =
+      message || `'${paramName || "value"}' cannot be null or undefined`;
+    throw new Error(message);
+  }
+};
+
+export const guardTypeOf = (value: any, type: string, message?: string) => {
```
mtarng, line 43: JSDocs, otherwise LGTM.

mtarng (approved):

bnookala:
I think you should make clear that this should be using for coercing and validating user inputs - as opposed to replacing typescript's type guards etc, but otherwise it looks ok

bnookala (approved):

src/lib/guard.ts
evanlouie, line 43: `value` needs be of type `number | string | boolean`.  `typeof` can't be used on `Array` or `Object` (`Array` extends `Object`, so `typeof` cant be confidently used).

If you want to support `number | string | boolean | Array | Object`, your code would have to look like:

```ts
export const guardTypeOf  = <T>(
  value: T,
  type: "number" | "string" | "boolean" | "array" | "object",
  message?: string
) => {
  if (
    ["number", "string", "boolean", "object"].includes(type) &&
    typeof value !== type
  ) {
    message =
      message || `${value} must be of type ${type}, ${typeof value} given`;
    throw new Error(message);
  } else if (type === "array" && Array.isArray(value) === false) {
    message =
      message || `${value} must be of type Array, ${typeof value} given`;
    throw new Error(message);
  }
};

```

evanlouie (changes requested):

src/commands/service/create.ts
```diff
@@ -41,26 +42,12 @@ export const createCommandDecorator = (command: commander.Command): void => {
...
-        if (typeof maintainerEmail !== "string") {
-          throw new Error(
-            `maintainerEmail must be of type 'string', ${typeof maintainerEmail} given.`
-          );
-        }
+        const stringType = "string";
+        guardTypeOf(serviceName, stringType);
+        guardTypeOf(packagesDir, stringType);
+        guardTypeOf(maintainerName, stringType);
+        guardTypeOf(maintainerEmail, stringType);
```
evanlouie, line 49: Does `guardTypeOf` actually act as a type guard for any of the variables?
Following lines 49, does the compiler actually know `serviceName`, `packagesDir`, etc... are of type `string`?
I don't believe the current setup will actually allow the compiler to know the types of the variables (ie; their still `any`) after the `guardTypeOf` calls; whereas the compiler knows their of type `string` using the `if (typeof maintainerEmail !== "string") {...` syntax.

evanlouie: Just played with this a bit, just as I feared

Fails:

```ts
const log = (msg: string) => console.log(msg);
const foo: unknown = "i'm a string";
guardTypeOf(foo, "string");
log(foo);
```

Passes:

```ts
const log = (msg: string) => console.log(msg);
const foo: unknown = "i'm a string";
if (typeof foo !== "string") {
  throw new Error("Expected string");
}
log(foo);
```

We should maintain to just use vanilla TS typeguards for now (`if (typeof ...`)).

mydiemho: I'm ok with using vanilla guards.  But is there a way to abstract it out to keep the main logic simple?

evanlouie: Sadly not yet. TS is planning on supporting `assert`ion guards which do exactly this in TS 3.7 https://devblogs.microsoft.com/typescript/announcing-typescript-3-7-beta/

But until then, the compiler isn't smart enough to really support a clean abstraction :(

PR #43, opened 2019-10-03
Adding global init command

samiyaakhtar:
Ability to run command `spk init -f /somewhere/file.yml`
- parses yml file and loads config into an object, by load any `{env:something}` references from `process.env`
- updated `spk deployment get` command to start using this config
- removed old `spk deployment init` command
- separating out onboard into its own command `spk deployment onboard`

Related to https://github.com/microsoft/bedrock/issues/625, https://github.com/microsoft/bedrock/issues/593


andrebriggs (approved):

config.env
```diff
@@ -0,0 +1,4 @@
+INTROSPECTION_STORAGE_ACCOUNT_NAME=
```
sarath-p: This file is not needed. We need to create two files. 
1) .env file with values that will be used running the app locally but not pushed to the repo with .gitignore since you don't want to share the values.
2) .env.example file will have all the variables without any values so others can create .env locally and configure with actual values.

spk-config.yml
```diff
@@ -0,0 +1,41 @@
...
+    tag:
+    repo_type: "Public"
+    private_repo_key:
+  azure:
+    az_cli:
+      azure_cli: "2.0.70"
+      Extensions:
+        aks_preview: "0.4.14"
+        # Support for Key vault / AzDO var groups / .env
+    az_auth:
```
sarath-p: az_auth will need to be removed here since they are configured with sensitive information in .env file.

src/commands/init.ts
```diff
@@ -0,0 +1,79 @@
+import commander from "commander";
+import dotenv = require("dotenv");
+import * as fs from "fs";
+import yaml from "js-yaml";
+import * as os from "os";
+import * as env from "../config";
+import { logger } from "../logger";
+import { Command } from "./command";
+
+const defaultFileLocation = os.homedir() + "/.spk-config.yml";
```
sarath-p: how about reading the filepath from .env?

src/config.ts
```diff
@@ -0,0 +1,8 @@
+export const introspectionStorageAccessKey: string = process.env
+  .INTROSPECTION_STORAGE_ACCESS_KEY!;
+export const introspectionStorageAccountName: string = process.env
```
sarath-p: Non-sensitive values can be configured directly in config.yaml. Sensitive values such as storage account key, service principal id should be configured in .env file.

sarath-p (changes requested):

config.env
samiyaakhtar: I think you're looking at a particular commit from this PR, because I removed this file later on? Still double checking to make sure

src/config.ts
samiyaakhtar: Yup agreed, I've removed this file already

src/commands/init.ts
samiyaakhtar: This is the default filepath, similar to kubeconfig, after they run init, we save the file at `~/.spk-config.yml` so that they don't have to keep passing us path to the config file for every command after init. This has been moved to another file

spk-config.yml
```diff
@@ -0,0 +1,45 @@
+infra:
```
evanlouie: Can we get a type definition for this file added to `src/types.d.ts`?

src/commands/init.ts
```diff
@@ -0,0 +1,111 @@
...
+  config = configObj;
+};
+
+/**
+ * Reads a YAML file and loads it into an object
+ * @param fileLocation path to the config file to read
+ */
+export const readYamlFile = (fileLocation: string): string => {
+  const data: string = fs.readFileSync(fileLocation, "utf-8");
+  const response = yaml.load(data);
```
evanlouie: should use `yaml.safeLoad` instead unless we need to eval JS code in yaml (https://github.com/nodeca/js-yaml#safeload-string---options-)

src/commands/init.ts
```diff
@@ -0,0 +1,111 @@
...
+  iterate(configObj);
+  // Set the global config so env vars are loaded into it
+  config = configObj;
+};
+
+/**
+ * Reads a YAML file and loads it into an object
+ * @param fileLocation path to the config file to read
+ */
+export const readYamlFile = (fileLocation: string): string => {
```
evanlouie: Since this is meant to deserialize the the config file, the return type should be the type def for the yaml config file

spk-config.yml
```diff
@@ -0,0 +1,45 @@
...
+    key: '${env:INTROSPECTION_STORAGE_ACCESS_KEY}' # Will be filled from keyvault later
+    partition_key: '${env:INTROSPECTION_STORAGE_PARTITION_KEY}' # Will be autofilled later
+  pipeline:
+    org: "epicstuff"
+    project: "hellobedrock"
+    access_token: "test-token"
+services:
+  project_name:
+  PAT:
+  Devops_org:
```
evanlouie: new line plz 🙏

config.example.env
```diff
@@ -0,0 +1,4 @@
+export INTROSPECTION_STORAGE_ACCOUNT_NAME=
+export INTROSPECTION_STORAGE_TABLE_NAME=
+export INTROSPECTION_STORAGE_PARTITION_KEY=
+export INTROSPECTION_STORAGE_PARTITION_KEY=
```
evanlouie: newline plz 🙏

spk-config.yml
evanlouie: Can you either change the extension to `yaml` or add `yml` to the `lint-staged` glob pattern in package.json?

src/commands/init.ts
```diff
@@ -0,0 +1,111 @@
...
+};
+
+/**
+ * Loads configuration from a given filename, if provided, otherwise
+ * uses the default file location ~/.spk-config.yml
+ * @param fileName file to load configuration from
+ */
+export const loadConfiguration = (fileName?: string) => {
+  if (!fileName) {
+    fileName = defaultFileLocation;
```
evanlouie, line 47: Is there a reason why we keep `defaultFileLocation` as a top level var instead of having it be the default in the function sig? (ie; `(filename: string = "os.homedir() + "/.spk-config.yml") => {...`)

src/commands/init.ts
evanlouie: If this is meant to be a general yaml reading function, can you turn it into a generic to support my first comment?:

```ts
export const readYamlFile = <T>(fileLocation: string): <T> => {
``` 

Then when you load the config file, you can pass it the type of the config file

evanlouie (changes requested):

src/commands/init.ts
```diff
@@ -0,0 +1,111 @@
...
+    logger.error(`File ${fileName} does not exist.`);
+    throw err;
+  }
+};
+
+/**
+ * Loads configuration from local env
+ * @param configYaml configuration in object form
+ */
+export const loadConfigurationFromLocalEnv = (configObj: any) => {
```
evanlouie: can you give more detail on what this is doing on the jsdoc?

evanlouie (changes requested):

src/commands/init.ts
samiyaakhtar: Because currently it's being exported into the test file, I'll look for alternatives

src/commands/init.ts
samiyaakhtar: Done

spk-config.yml
samiyaakhtar: Changed to yaml

src/commands/init.ts
samiyaakhtar: Kept this since I'm using it to unit test

src/commands/init.ts
samiyaakhtar: Done

evanlouie (approved):

sarath-p (approved):

PR #52, opened 2019-10-07
Refactoring config file to account for common config

### LINKED ISSUE #655: Use spk-config in azurecredentials
**As a:** developer

**I want:** to use the spk-config file

**So that:** azurecredentials doesn't have to rely on the config.ts file

**Describe the solution you'd like:** delete config.ts and refactor azurecredentials.ts to use values from `spk-config.yaml`

**Does this require updates to documentation?:** no

samiyaakhtar:
Moved some things to common areas, which belong under specific AzDO or AzDO pipelines sections. 

Closes https://github.com/microsoft/bedrock/issues/655

### DIFF
--- a/spk-config.yaml
+++ b/spk-config.yaml
@@ -1,45 +1,52 @@
-infra:
-  terraform_check: true
-  git_check: true
-  helm_check: true
-  az_cli_check: true
-  env_var_check: true
-  terraform: "v0.12.6"
-  helm: "v2.14.3"
-  git: "2.17.1"
-  bedrock:
-    # Releases are through tags / branches
-    source: "git@github.com:microsoft/bedrock.git"
-    tag:
-    repo_type: "Public"
-    private_repo_key:
-  azure:
-    az_cli:
-      azure_cli: "2.0.70"
-      Extensions:
-        aks_preview: "0.4.14"
-        # Support for Key vault / AzDO var groups / .env
-    az_auth:
-      service_principal_id: "client-id"
-      service_principal_secret: "client-secret"
-      subscription_id: "subscription-id"
-      tenant_id: "tenant-id"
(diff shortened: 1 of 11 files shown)


spk-config.yaml
```diff
@@ -1,45 +1,52 @@
...
+  tenant_id: "tenant-id"
+
+# Azure CLI Specific Configuration
+azure-cli:
+  version: "2.0.70"
+  Extensions:
+    aks-preview: "0.4.14"
+
+# Azure DevOps Configuration
+azure-devops: # Should this live under the `spk` configuration?
```
bnookala: I think we may need to enumerate the other repositories here as well (ie hld, materialized manifest)

bnookala (changes requested):

spk-config.yaml
samiyaakhtar: You mean for the case when hld, manifest may be under a different azure org and project? cc: @mtarng

bnookala: no, we just need it for creating pipelines between them - let's assume one project for now

yradsmikham (approved):

edaena (approved):

spk-config.yaml
```diff
@@ -1,45 +1,35 @@
+# Azure CLI Specific Configuration
+azure_cli:
+  version: "2.0.70"
+  Extensions:
+    aks_preview: "0.4.14"
+
+# Azure DevOps Configuration
+azure_devops: # Should this live under the `spk` configuration?
+  org: "my-az-do-org"
+  project: "my-mono-repo-project"
```
bnookala, line 10: i think you can remove this, since it's in a separate config

bnookala (approved):

PR #53, opened 2019-10-07
Bedrock 597: Make sure all configuration pre-reqs are present

edaena:
Make sure service introspection configuration is set.

Related to https://github.com/microsoft/bedrock/issues/597


src/commands/mocks/spk-config-deployment.yaml
```diff
@@ -0,0 +1,40 @@
...
+      subscription_id: "subscription-id"
+      tenant_id: "tenant-id"
+    az_backend:
+      resource_group_name: "StorageAccount-ResourceGroup"
+      storage_account_name: "abcd1234"
+      container_name: "tf-state"
+      key: "prod.terraform.tf-state"
+deployment:
+  storage:
+    account_name: "my_storage_account"
```
samiyaakhtar: should we just utilize the mocks file we already have, `mocks/spk-config.yaml`? That way we're not modifying multiple files every time configs change

src/commands/deployment/validate.ts
```diff
@@ -0,0 +1,66 @@
...
+      }
+    }
+  }
+
+  if (missingConfig != "") {
+    logger.error("Validation failed. Missing configuration: " + missingConfig);
+    return false;
+  }
+
+  return true;
```
samiyaakhtar, line 66: Maybe we should print something if the validation is successful?

edaena: good point...

src/commands/deployment/validate.ts
```diff
@@ -0,0 +1,67 @@
...
+    .action(() => {
+      isValidConfig();
+    });
+};
+
+/**
+ * Validates that the deployment configuration is specified.
+ */
+export const isValidConfig = (): boolean => {
+  let missingConfig = "";
```
bnookala: instead of appending (and re-binding) to the same string for missing configs, could you use an array instead? ie

```
const missingConfig = [];

if (!config.deployment) {
  missingConfig.push("deployment")
} 

// other checks...

const missingConfigsMessage = missingConfig.join(' ');
...
```

bnookala (changes requested):

samiyaakhtar (approved):

bnookala (approved):

PR #55, opened 2019-10-08
Add pipeline command

bnookala:
This PR:

- Add pipeline creation command
- Add pipeline creation unit tests
- Add logging lines to pipelines lib

This PR:
- Does not use global config (yet)

### DIFF
--- a/src/commands/service/pipeline.ts
+++ b/src/commands/service/pipeline.ts
@@ -1,0 +1,187 @@
+import fs from "fs";
+import yaml from "js-yaml";
+import path from "path";
+
+import commander = require("commander");
+import { logger } from "../../logger";
+
+import { IBuildApi } from "azure-devops-node-api/BuildApi";
+import {
+  createPipelineForDefinition,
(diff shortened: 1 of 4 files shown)


src/commands/service/pipeline.ts
```diff
@@ -0,0 +1,187 @@
...
+        repoName,
+        repoUrl,
+        devopsProject,
+        projectPath
+      } = opts;
+
+      try {
+        if (typeof pipelineName !== "string") {
+          throw new Error(
+            `pipelineName must be of type 'string', ${typeof pipelineName} given.`
```
evanlouie: Should we reference the variable name or the flag that populates it (ie; `--pipeline-name`)?

bnookala: oh yeah that might make more sense, though there's a number of places presently where we actually use the internal variable name…

bnookala: Fixed

mtarng (approved):

evanlouie (approved):

PR #56, opened 2019-10-08
adding flag to commit and push new service changes

mtarng:
Closes out part 1 of https://github.com/microsoft/bedrock/issues/577


bnookala:
Your builds are failing (or being cancelled) because of a typing error with a module: 
```
TS7016: Could not find a declaration file for module 'git-url-parse'. '/home/runner/work/spk/spk/node_modules/git-url-parse/lib/index.js' implicitly has an 'any'
```

bnookala:
I *think* you can fix this issue by installing the typings for the module via: `yarn add --save @types/git-url-parse`

src/lib/gitutils.ts
```diff
@@ -0,0 +1,130 @@
...
+    logger.warn(`unable to checkout git branch ${branchName}.`);
+  }
+};
+
+/**
+ * Create and checkout a new branch.
+ *
+ * @param branchName
+ */
+export const checkoutNewBranch = async (branchName: string) => {
```
bnookala: could this be combined with `checkoutBranch`, with some kind of option to create a new branch?

src/lib/gitutils.ts
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * Checkout the given branch.
+ *
+ * @param branchName
+ */
+export const checkoutBranch = async (branchName: string) => {
+  try {
+    await exec("git", ["checkout", `${branchName}`]);
+  } catch (_) {
+    logger.warn(`unable to checkout git branch ${branchName}.`);
```
bnookala: you may want to throw these errors instead of suppressing them, for debuggability. you should do this here, and everywhere else you're catching and logging warnings

src/lib/gitutils.ts
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Gets the current working branch.
+ */
+export const getCurrentBranch = async () => {
+  try {
+    const branch = await exec("git", ["rev-parse", "--abbrev-ref", "HEAD"]);
+    return branch;
+  } catch (_) {
+    logger.warn("Unable to determine current branch.");
```
bnookala: use logger.error for these

src/commands/service/create.ts
```diff
@@ -241,4 +263,18 @@ export const createService = async (
...
+    await commitDir(newServiceDir, serviceName);
+    await pushBranch(serviceName);
+    await checkoutBranch(currentBranch);
+    await deleteBranch(serviceName);
+    const pullRequestLink = await getPullRequestLink(
+      currentBranch,
+      serviceName,
+      await getOriginUrl()
+    );
+    logger.info(`Link to create PR: ${pullRequestLink}`);
```
bnookala: what happens if any of these should fail?

bnookala (changes requested):

bnookala (approved):

evanlouie (approved):

PR #58, opened 2019-10-09
Integrate infra init as part of global init

### LINKED ISSUE #615: As a developer, I want to initialize infrastructure for Bedrock and host the state of such environment.
**As a:**
Developer

**I want:**
a CLI to store validation against a local environment to ensure all prerequisites are installed and authentication (to Azure) is valid.

**So that:**
The CLI is able to identify missing requirements for infrastructure deployment.

The CLI command should include the following functionality:
- [x] Rename `spk infra init` to `spk infra validate`
- [x] Write to `spk-config.yaml` file. For example:
```
infra:
  checks:
      terraform: true
      git: true
      az: true
      helm: true
      az_login_check: true
      env_var_check: true
  versions:
      terraform: "v0.12.6"
      helm: "v2.14.3"
      git: "2.17.1"
```
- [x] Handle "soft checks" and "hard checks". When the global `init` executes, an infra "soft check" will occur, when future `spk infra` commands execute, an infra "hard check" will occur.

Acceptance Criteria:

spk tool (CLI) will write to `spk-config.yaml`, and perform soft/hard checks.

Yes

yradsmikham:
Closes https://github.com/microsoft/bedrock/issues/615.


samiyaakhtar (approved):

src/commands/init.ts
```diff
@@ -29,6 +34,20 @@ export const initCommandDecorator = (command: commander.Command): void => {
         }
         loadConfiguration(opts.file);
 
+        await validatePrereqs(["terraform", "git", "az", "helm"], true);
+
+        await validateAzure(true);
+
+        await validateEnvVariables(
+          [
+            "ARM_SUBSCRIPTION_ID",
```
NathanielRose, line 43: Just a note based on what we discussed yesterday, if we're using static env vars for infra validation, should we not be expecting the same env vars in spk.config?

NathanielRose (approved):
LGTM

src/commands/init.ts
yradsmikham: That is still an open discussion. I think for now, we can assume we expect these environment variables to be defined, but if we do not have a reason for it to be stored in the `spk-config.yaml`, we should leave it out. This can obviously change!

PR #61, opened 2019-10-10
Fixes `fetch` issue running `onboard` command from `spk`

### LINKED ISSUE #658: onboard command throwing an error `fetch is not a function` running from spk
**Describe the bug:**

`spk onboard` command throwing an error `fetch is not a function` running from  but works from `ts-node onboard`

**To Reproduce:**
run `spk-macos deployment onboard -n testname -l westus2 -r test-rg -k testvault`

**Expected behavior:**
Check whether the storage account exist in the resource group. If not, creates a new storage account.

**Screenshots:**

**Additional context:**

No

sarath-p:
closes  microsoft/bedrock#658. 

Removed wrapper new Error object in all catch blocks in Storage.ts file.

Added null/undefined check for accounts variable to handle the scenario when no storage accounts exist in the resource group.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
   "devDependencies": {
     "@types/cli-table": "^0.3.0",
     "@types/dotenv": "^6.1.1",
+    "@types/isomorphic-fetch": "^0.0.35",
     "@types/jest": "^24.0.18",
     "@types/jest-when": "^2.7.0",
     "@types/js-yaml": "^3.12.1",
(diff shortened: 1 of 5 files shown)


edaena (approved):

package.json
```diff
@@ -65,7 +66,9 @@
     "cli-table": "^0.3.1",
     "commander": "^3.0.1",
     "dotenv": "^8.1.0",
+    "es6-promise": "^4.2.8",
```
evanlouie: Remove this (`yarn remove es6-promise`). We shouldn't need this as all nodejs versions come with promise support

evanlouie (changes requested):

samiyaakhtar (approved):

src/index.ts
```diff
@@ -11,6 +11,7 @@
  * - if (typeof var === 'number') {}
  * - etc...
  */
+import "isomorphic-fetch";
```
samiyaakhtar, line 15: might want to add a comment explaining why this is being imported, when it's not being used

samiyaakhtar (changes requested):

samiyaakhtar (approved):

package.json
sarath-p: Good catch, removed it!

evanlouie (approved):

PR #62, opened 2019-10-11
Create Service Revision

evanlouie:
- Added ring tracking to bedrock.yaml
- Added command to create PR against all default rings in bedrock.yaml


src/commands/service/create-revision.ts
```diff
@@ -0,0 +1,176 @@
...
+      "Source branch to create the pull request from; defaults to the current branch"
+    )
+    .option("-t, --title <title>", "Title of the pull request; not required")
+    .option(
+      "-d, --description <description>",
+      "Description of the pull request; not required"
+    )
+    .option(
+      "--remote-url <remote-url>",
+      "The remote host to create the pull request in; defaults to the URL for 'origin'; defaults to the URL for 'origin'"
```
bnookala: text here is duped

src/commands/service/create-revision.ts
```diff
@@ -0,0 +1,176 @@
...
+            }
+            if (stdout.trim().length === 0) {
+              throw Error(
+                `Unable to parse current branch from git client. Ensure '${gitBranchCmd}' returns a proper response`
+              );
+            }
+            logger.info(
+              `Successfully parsed current branch from git client; --source-branch for pull request set to '${stdout.trim()}'`
+            );
+            sourceBranch = stdout.trim();
```
bnookala: perhaps this can be pulled into git utils? https://github.com/CatalystCode/spk/blob/master/src/lib/gitutils.ts

src/commands/service/create-revision.ts
```diff
@@ -0,0 +1,176 @@
...
+              if (stderr.length > 0) {
+                logger.warn(
+                  `Non-zero length stderr outputted from '${gitOriginRemoteCmd}': '${stderr}'`
+                );
+              }
+              return stdout.trim();
+            }
+          );
+          logger.info(`Parsed remote-url for origin: '${originRemote}'`);
+          remoteUrl = originRemote;
```
bnookala: add this to git utils? https://github.com/CatalystCode/spk/blob/master/src/lib/gitutils.ts

src/commands/service/create-revision.ts
```diff
@@ -0,0 +1,176 @@
...
+        ////////////////////////////////////////////////////////////////////////
+        // Main
+        ////////////////////////////////////////////////////////////////////////
+        // Make a PR against all default rings
+        for (const ring of defaultRings) {
+          if (typeof title !== "string") {
+            title = `[SPK] ${sourceBranch} => ${ring.branch}`;
+            logger.info(`--title not set, defaulting to: '${title}'`);
+          }
+          await createPullRequest(title, sourceBranch, ring.branch, {
```
bnookala, line 145: I think we already have code that does… once again in gitutils https://github.com/CatalystCode/spk/blob/master/src/lib/gitutils.ts#L109

bnookala (approved):
i think im fine with merging this in - my only concern is that it _looks_ like we have a lot of this code duplicated? maybe this is something we can address some other time

src/types.d.ts
```diff
@@ -41,6 +41,11 @@ export interface IHelmConfig {
  * Used to capture service meta-information regarding how to deploy
  */
 export interface IBedrockFile {
+  rings: {
```
mtarng: Shouldn't this be an optional field?

evanlouie: Marked as optional

src/commands/service/create-revision.ts
evanlouie: Different function, that one just generates the link for a PR. This actually uses the azdo library to create the pull request

src/commands/service/create-revision.ts
evanlouie: Function was actually already in there. Refactored to use that one instead

mtarng (approved):

PR #68, opened 2019-10-15
add script to bump package version

mydiemho:
1. Developer can use this script to update `package.json` and push to a new branch
1. Once this branch is merge, they can manually 
    * run `yarn build` to generate the `dist` directory with the 3 binaries
    * manually create a release on GitHub

Related to https://github.com/microsoft/bedrock/issues/651

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "spk",
-  "version": "1.0.0",
+  "version": "1.0.0-0",
   "description": "The missing Bedrock CLI",
   "author": "Evan Louie <evan.louie@microsoft.com> (https://evanlouie.com)",
   "contributors": [],
--- a/scripts/prerelease-version-bump.sh
+++ b/scripts/prerelease-version-bump.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+RELEASE_BRANCH=$(whoami)/release
+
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "spk",
-  "version": "1.0.0",
+  "version": "1.0.0-0",
```
bnookala, line 3: We should start this from 0.1.0

bnookala (changes requested):
@mydiemho per https://github.com/microsoft/bedrock/issues/651#issuecomment-541832442, I wrote i added some requirements - would like to see those addressed.

package.json
mydiemho: @bnookala Depends on how you want to version this, there are 2 ways.  I personally prefer option 1, as you're working toward 1.0.0 anyway

## Option 1

Using `1.0.0-0`, you can bump up each pre-release like this
* 1.0.0-1
* 1.0.0-2

and when you're ready, doing a non pre-release, you'll get 1.0.0

## Option 2

Starting at 0.1.0, you're probably thinking of going from
* 0.1.0 => 0.2.0 => 0.3.0

mydiemho:
will open new pr with pipeline

PR #69, opened 2019-10-15
Add template to cluster definition

### LINKED ISSUE #667: Add "template" key to the definition.json as part of `spk infra scaffold`
**Describe the bug:**
The "template" key needs to be added as part of the `definition.json` when `spk infra scaffold` executes.

**To Reproduce:**
Execute `spk infra scaffold`

**Expected behavior:**
```
{​
    name: "discovery-service",

    source: "https://github.com/fabrikam/bedrock",
    template: "cluster/environments/fabrikam-common-infra",
    version: "d7d905e6551",

    backend: {​
        storage_account_name: "terraformstate"
        container_name: "discoveryservice"
    }​,

    variables: {​
        agent_vm_count: 3,
        gitops_ssh_url: "git@github.com:fabrikam/discovery-cluster-manifests.git"
    }​
}
```

Yes

yradsmikham:
- Added `template` as part of definition.json 
- `--template` will now take in a path as opposed to `variables.tf` file.
- Added helper functions (`copyTfTemplate, renameTfvars) to prepare for Terragrunt integration.
- Renamed `validate` function to match other infra commands
- Updated README

Closes https://github.com/microsoft/bedrock/issues/667.


src/commands/infra/scaffold.ts
```diff
@@ -60,8 +63,53 @@ export const validateVariablesTf = async (
...
+  }
+  return true;
+};
+
+/**
+ * Rename any .tfvars file by appending ".backup"
+ *
+ * @param dir path to template directory
+ */
+export const renameTfvars = async (dir: string): Promise<void> => {
```
jmspring, line 81: Rather than copying then renaming TFVar files, why not just avoid copying them in the first place?

yarn.lock
```diff
@@ -522,6 +522,13 @@
   resolved "https://registry.yarnpkg.com/@types/events/-/events-3.0.0.tgz#2862f3f58a9a7f7c3e78d79f130dd4d71c25c2a7"
```
jmspring, line 522: Is this a file we should really be checking in?

jmspring (commented):
Additionally are there any updates/changes to the tests?

yarn.lock
yradsmikham: Yes, it should be commited!

src/commands/infra/scaffold.ts
yradsmikham: fs library has a method that allows to copy an entire directory without having to write logic that involves iteration. I feel that it is best to include the tfvars file as part of the copy process, instead of removing or excluding it because it gives users the option to toggle between using Terraform or Terragrunt if that becomes necessary. In the case, where a user wants to use Terraform, they would just need to rename the tfvars file (remove ".backup" from the name). Do you think we should not provide this option?

src/commands/infra/README.md
```diff
@@ -60,6 +59,8 @@ It will do the following:
 Usage:
 spk infra scaffold|s [options]
 
+> `spk infra scaffold --name discovery-service --source https://github.com/microsoft/bedrock --version "0.0.1" --template .bedrock/cluster/environments/azure-simple`
```
NathanielRose: Switch template example to use full path

src/commands/infra/README.md
```diff
@@ -60,6 +59,8 @@ It will do the following:
...
+=======
```
NathanielRose: Remove

src/commands/infra/scaffold.ts
NathanielRose: I think the `tfvar.backup` is a good approach. We should store it just in case the user wishes to pivot from leveraging terragrunt and needs to debug directly with terraform.

NathanielRose (commented):
Bringing this thought up again- maybe we should be storing base terraform templates created through scaffold in an `.infrastructure/` folder at the root for organization and to be added to .gitignores .

yradsmikham:
> Additionally are there any updates/changes to the tests?

@jmspring, no updates made to the test because I am simply using libraries to write the helper functions. I did not see a need in unit testing fs library commands (e.g. `rename`, `copy`).  Let me know if you feel differently.

yradsmikham:
> Bringing this thought up again- maybe we should be storing base terraform templates created through scaffold in an `.infrastructure/` folder at the root for organization and to be added to .gitignores .

This seems more like a Day 2 item, and I do agree that it needs to be addressed somehow, possibly something we handle outside of scaffolding. Let's have this flushed out in an issue first.

NathanielRose (approved):

jmspring (approved):
lgtm

PR #73, opened 2019-10-16
Creates new Variablegroup with a link to Azure Key Vault

### LINKED ISSUE #649: Create a Variable Group with a link to Azure KeyVault
**As a:**

Developer/Operator

**I want:**

To be able to create a variable group in AzDo project with bedrock CLI. 

I expect my CLI to:
  - [x] Ability to create Variable Group in AzDo project from the configuration
     - [x] Create Variable Group with `Vsts` type
        - Map configured key/value pairs as Variables in the group.
     - [x] Create Variable Group with `AzureKeyVault` type
        - Map configured Azure Key Vault Secrets in Variable Group
        - Create Service Connection EndPoint to authenticate with Key Vault if Service Connection does not exist with the name in the payload.
- [x] Accept variable group payload from the command without depending on `spk` config including Service Creation payload for authorization with Key Vault
- [x] Create unit test cases

**So that:**

I can easily setup Bedrock GitOps features and using secrets from Azure Key Vault with out a manual process.

An option in CLI tool to create a Variable Group in AzDo project.

Manual creation in AzDo UI

**Acceptance Criteria:**

An AzDo project contain newly created variable group with a link to Azure Key Vault for secrets. 

Yes

sarath-p:
Ability to create two types of Variable Groups in AzDo project based on the configuration in spk-config.yaml

1. Vsts type
   - Enables "access to all pipelines" flag
   - Created variables based on the key/value pairs from the configuration

2. AzureKeyVault type
   - Enables "access to all pipelines" flag
   - Enables Key Vault linking
   - Creates a new Service Connection if existing service connection is not found with a configured Service Connection Name
    - Adds variables based on the Key Vault secret names configured as variables in the Variable Group

Closes https://github.com/microsoft/bedrock/issues/649


.env.example
```diff
@@ -3,3 +3,10 @@ AZURE_TENANT_ID="AAD tenant id"
 AZURE_CLIENT_ID="Azure service principal client Id"
 AZURE_CLIENT_SECRET="Azure service principal client secret/password"
 AZURE_SUBSCRIPTION_ID="Azure subscription id" 
+
+AZDO_PERSONAL_ACCESS_TOKEN=" Azure DevOps Personal Access Token"
```
samiyaakhtar: this should be read from config, `config.azure_devops.access_token`

spk-config.yaml
```diff
@@ -6,13 +6,33 @@ azure_cli:
...
+    key_vault_provider:
+      name: "sarathpvault" # name of the Azure Key Vaukt with Secrets
+      secrets: ["secret1", "secret2"] # Names of secrets from Azure Key Vault to map in Variable Group
+      service_connection: # service connection is required to authorize with Azure Key Vault
+        name: "Bedrock-GitOps-KeyVault" # if the service connection with this name does not exist, the following values are required to create a new service connection with this name
+        subscription_id: "${env:AZDO_SERVICE_CONNECTION_SUBSCRIPTION_ID}" # Azure Subscription id where Key Vault exist
+        subscription_name: "CSE Dev Crews Americas West" # Azure Subscription name where Key Vault exist
+        service_principal_id: "${env:AZDO_SERVICE_CONNECTION_CLIENT_ID}" # Service Principal Id that has 'Get' and 'List' in Key Vault Access Policy
+        service_principal_secret: "${env:AZDO_SERVICE_CONNECTION_CLIENT_SECRET}" # Service Principal secret for the above Service Principal Id
+        tenant_id: "${env:AZDO_SERVICE_CONNECTION_TENANT_ID}" # AAD Tenant Id for the above Service Principal
```
samiyaakhtar: Do these need to be in `spk-config.yaml`?

yradsmikham: Is there a need to have all of this be part of our sample `spk-config.yaml`? I know there was a discussion on minimizing the content of `spk-config.yaml`, and if there is no specific reason to have it be stored here, then we should remove it. An alternative would be to document some of this stuff in a README.

package.json
```diff
@@ -73,6 +73,7 @@
     "node-emoji": "^1.10.0",
     "shelljs": "^0.8.3",
     "spektate": "^0.1.5",
+    "ts-node": "^8.4.1",
```
samiyaakhtar: May not need this

spk-config.yaml
yradsmikham: same comment as @samiyaakhtar, haha.

sarath-p: @yradsmikham, @samiyaakhtar - any ideas how to get this information in the runtime without passing these as arguments from a command? Too many arguments to pass from the command.

package.json
sarath-p: I will take a look since it was added from debugging in VS Code.

spk-config.yaml
samiyaakhtar: These would be needed only during the onboard command, right?

sarath-p: The actual command is not yet finalized but won't be very frequent similar to the onbaord command.

src/lib/pipelines/azDo.ts
```diff
@@ -0,0 +1,38 @@
+import { getPersonalAccessTokenHandler, WebApi } from "azure-devops-node-api";
+import { ITaskAgentApi } from "azure-devops-node-api/TaskAgentApi";
+import { RestClient } from "typed-rest-client";
+import { logger } from "../../logger";
+
```
edaena: Missing documentation for each function.

src/lib/pipelines/serviceConnection.test.ts
```diff
@@ -0,0 +1,104 @@
...
+import { IServiceConnectionConfiguration } from "../../types";
+import * as serviceConnection from "./serviceConnection";
+
+jest.spyOn(serviceConnection, "getServiceConnectionByName").mockImplementation(
+  async (): Promise<any> => {
+    return undefined;
+  }
+);
+
+// jest.spyOn(serviceConnection, "createServiceEndPointData").mockImplementation(
```
edaena: Remove commented code

src/lib/pipelines/serviceConnection.test.ts
```diff
@@ -0,0 +1,104 @@
...
+//     }
+//     const ret: any = {};
+//     return ret as JSON;
+//   }
+// );
+
+// let serviceConnectionParams: JSON;
+
+beforeEach(() => {
+  // const config = getConfig();
```
edaena: remove commented code

spk-config.yaml
```diff
@@ -5,7 +5,26 @@ azure_devops: # Should this live under the `spk` configuration?
...
+    key_vault_provider:
+      name: "mykeyvault" # name of the Azure Key Vaukt with Secrets
+      secrets: ["secret1", "secret2"] # Names of secrets from Azure Key Vault to map in Variable Group
+      service_endpoint: # service endpoint is required to authorize with Azure Key Vault
+        name: "Bedrock-GitOps-KeyVault" # if the service endpoint with this name does not exist, the following values are required to create a new service connection with this name
+        subscription_id: "subscription-id" # Azure Subscription id where Key Vault exist
+        subscription_name: "subscription-name" # Azure Subscription name where Key Vault exist
+        service_principal_id: "service-principal-id" # Service Principal Id that has 'Get' and 'List' in Key Vault Access Policy
+        service_principal_secret: "service-principal-secret" # Service Principal secret for the above Service Principal Id
+        tenant_id: "tenant-id" # AAD Tenant Id for the above Service Principal
```
samiyaakhtar: if the name does not exist, I'm thinking all these values would be needed either as a one time, or we could reuse them from under `introspection.azure` in config.

spk-config.yaml
```diff
@@ -5,7 +5,26 @@ azure_devops: # Should this live under the `spk` configuration?
...
+  variable_group:
+    name: "myvg"
+    description: "myvg description"
+    variables: # Either this or Key Value provider should be configured for this variable group
+      var1:
+        value: value1
+        isSecret: true
+      var2:
+        value: value2
+        isSecret: false
```
samiyaakhtar: do you think it makes sense to prompt the user to enter values for their variables (or take them via command line arguments), instead of saying var1 and var2, which does not scale in case they want 10 variables in their VG?

src/types.d.ts
```diff
@@ -97,15 +99,45 @@ export interface IAzurePipelinesYaml {
...
+  };
+}
+
 export interface IConfigYaml {
+  azure_cli?: {
+    version?: string;
+    Extensions?: {
+      aks_preview?: string;
+    };
+  };
```
samiyaakhtar: pull from master, this has been removed i believe

spk-config.yaml
edaena: I agree about prompting command line arguments for one time use only values. I know the infra folks were following this flow. spk-config holds values for things that are needed repeatedly.

package.json
evanlouie: Move to devDependencies please

src/commands/deployment/onboard.ts
```diff
@@ -35,6 +36,7 @@ export const onboardCommandDecorator = (command: commander.Command): void => {
     )
     .action(async opts => {
       try {
+        await addVariableGroupWithKeyVaultMap();
```
samiyaakhtar: should we make this part optional, or add a command line argument, in case they don't want to create a variable group when onboarding? Right now, it seems like this is a mandatory step within `onboard`

samiyaakhtar (changes requested):
Based on our discussion, let's move the config outside of spk-config.yaml and add it to README 👍

spk-config.yaml
sarath-p: The Service Principal for authorizing with Key Vault can be different with minimum `get` and `list` permissions in Key Vault access policy instead of `owner` permissions requirement in `introspection` area.

src/lib/pipelines/serviceConnection.test.ts
sarath-p: will be removed in final commit.

spk-config.yaml
sarath-p: Prompts for each command also has scalability issues i.e. 10 prompts for 10 variables. I will enable accepting a payload form the file with other required variables.

src/commands/deployment/onboard.ts
sarath-p: It was only for testing and will removed in the final commit.

docs/variable-group.md
```diff
@@ -0,0 +1,121 @@
+# Variable Group
+
+Create
+[variable group](https://docs.microsoft.com/en-us/azure/devops/pipelines/library/variable-groups)
+in Azure DevOps project.
+
+Usage:
+
+```
+spk service [command] [options]
```
andrebriggs: @sarath-p you mean `spk variable-group`?

sarath-p: yes, thanks for catching copy/paste error!

andrebriggs (approved):

src/commands/variable-group/create.ts
```diff
@@ -0,0 +1,127 @@
...
+            `--project must be of type 'string', ${typeof project} specified.`
+          );
+        }
+
+        if (
+          typeof personalAccessToken !== "undefined" &&
+          typeof personalAccessToken !== "string"
+        ) {
+          throw Error(
+            `--personal-access-token must be of type 'string', ${typeof personalAccessToken} specified.`
```
samiyaakhtar, line 72: Shouldn't this be falling back to use personal access token from config, when this is not specified? Or, we remove the comment "falls back to azure_devops.access_token in spk config"

Same for project and orgName

src/types.d.ts
```diff
@@ -97,15 +100,44 @@ export interface IAzurePipelinesYaml {
...
```
samiyaakhtar: I believe this is not needed?

samiyaakhtar (commented):
LGTM except few nit comments above

src/commands/variable-group/create.ts
sarath-p: that is correct. It is only checking for the type of "string" when the argument is specified.

src/types.d.ts
sarath-p: No, git didn't pick up changes form the master for some reason so had to do force merge to overwrite.

samiyaakhtar (approved):

PR #74, opened 2019-10-16
Revised infra block for spk-config.yaml

yradsmikham:


andrebriggs (approved):

src/types.d.ts
```diff
@@ -102,11 +102,6 @@ export interface IConfigYaml {
     terraform?: string;
     helm?: string;
     git?: string;
-    bedrock?: {
-      source?: string;
-      tag?: string;
-      access_token?: string;
-    };
   };
```
samiyaakhtar: Infra is not using this portion either, right? 
```
  azure_cli?: {
    version?: string;
    Extensions?: {
      aks_preview?: string;
    };
  };

```

yradsmikham: That's correct. Right now, we are not using it. Let's remove it!

samiyaakhtar (approved):
:shipit:

PR #91, opened 2019-10-21
Added a top level test to demonstrate in-code end-to-end test

evanlouie:
Added a top level test to demonstrate how to test a top level command.
To do so, this test spawns its own ts-node repl and executes the code as
if it were running the spk executable itself.

These tests do not add to the code coverage exported from Jest, but are
good for testing end-to-end scenarios; With this, we can programmatically
test the execution of sequentially top level commands and ensure the
output of the commands is correct.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "shx": "^0.3.2",
     "ts-jest": "^24.0.2",
     "ts-loader": "^6.0.4",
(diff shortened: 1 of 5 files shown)


andrebriggs:
@evanlouie Going to close this since we have a way to do e2e testing in this fashion.

PR #97, opened 2019-10-23
Adding config to service create-pipeline

### LINKED ISSUE #676: [SPK] spk service create-pipeline should read from global config
**As a:**
SPK user 

**I want:**
`spk service create-pipeline` to be updated with the global configs

**So that:**
I don't need to use flags/options for already fields in the configured the spk config file.

SPK service create-pipeline _should_:

- [ ] Read all necessary config values from the global config settings

none

**Additional context:**

All flags are presently passed along the command line, this should be changed.

Yes!

mtarng:
option flags will override spk config values.

Closes: https://github.com/microsoft/bedrock/issues/676

### DIFF
--- a/src/commands/service/pipeline.ts
+++ b/src/commands/service/pipeline.ts
@@ -1,13 +1,14 @@
-import commander = require("commander");
-import { logger } from "../../logger";
-
 import { IBuildApi } from "azure-devops-node-api/BuildApi";
+import commander = require("commander");
+import path from "path";
+import { Config } from "../../config";
 import {
   createPipelineForDefinition,
   definitionForAzureRepoPipeline,
   getBuildApiClient,
   queueBuild
 } from "../../lib/pipelines/pipelines";
+import { logger } from "../../logger";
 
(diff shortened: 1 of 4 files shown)


src/commands/service/pipeline.ts
```diff
@@ -149,7 +170,7 @@ export const installPipeline = async (
     repositoryName: repoName,
     repositoryUrl: repoUrl,
     yamlFileBranch: "master",
-    yamlFilePath: `packages/${serviceName}/azure-pipelines.yaml`
+    yamlFilePath: path.join(packagesDir, serviceName, "azure-pipelines.yaml") // This may not work if we're using a non-mono repository and azure-pipelines.yaml is in the root directory.
```
mtarng, line 173: Need to check this functionality based on where azure-pipelines.yaml is located...

@evanlouie @bnookala 

This probably goes into our `spk project init` rewrite and how we'll onboard services in the future.

bnookala (approved):

evanlouie (approved):

PR #98, opened 2019-10-23
Fix broken `variables` type for azure-pipelines.yaml

evanlouie:
Fixed an issue with the starter azure-pipelines.yaml an top level type
definition for it. `variables` expects a sequence of maps but a top
level map was being generated instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "@types/jest": "^24.0.18",
     "@types/jest-when": "^2.7.0",
     "@types/js-yaml": "^3.12.1",
-    "@types/mock-fs": "^3.6.30",
+    "@types/mock-fs": "^4.10.0",
     "@types/node": "^12.7.8",
     "@types/node-emoji": "^1.8.1",
     "@types/shelljs": "^0.8.5",
@@ -30,7 +30,7 @@
     "jest": "^24.9.0",
     "jest-when": "^2.7.0",
     "lint-staged": ">=8",
-    "mock-fs": "^4.10.1",
+    "mock-fs": "^4.10.2",
     "nyc": "^14.1.1",
     "pkg": "^4.4.0",
     "prettier": "^1.18.2",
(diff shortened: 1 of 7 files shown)


mtarng (approved):

mtarng (changes requested):
We should have a fileutils test for`generateStarterAzurePipelinesYaml`

evanlouie:
what would it test? how would we test it? We've been avoiding doing network calls to azdo. So unless we actually do that, theirs no way to validate the yaml is valid

package.json
```diff
@@ -60,7 +60,7 @@
   },
   "dependencies": {
     "@azure/arm-storage": "^10.1.0",
-    "@azure/keyvault-secrets": "^4.0.0-preview.5",
+    "@azure/keyvault-secrets": "4.0.0-preview.5",
```
evanlouie, line 63: @sarath-p note I've pinned this at `preview.5`. They seem to change their module structure in newer versions and breaks our code.

evanlouie:
@mtarng i've added a test for the starter yaml

src/lib/fileutils.test.ts
```diff
@@ -236,3 +253,73 @@ describe("generating service Dockerfile", () => {
     );
   });
 });
+
+describe("starterAzurePipelines", () => {
+  // Create a random workspace dir before every test
+  let randomDirPath = "";
+  beforeEach(() => {
+    randomDirPath = path.join(os.tmpdir(), uuid());
```
evanlouie, line 261: note: i've had to use manual fs creation as opposed to `mock-fs` because it doesn't play nicely with `path.resolve` which `generateStarterAzurePipelinesYaml` uses

bnookala (approved):

mtarng (approved):

package.json
sarath-p: Sounds good. We can unpin when they release a non=preview version.

PR #99, opened 2019-10-23
Clean up spk dashboard

edaena:


src/commands/deployment/dashboard.test.ts
```diff
@@ -45,3 +45,30 @@ describe("Validate dashboard container pull", () => {
...
+        const dashboardContainerId2 = await launchDashboard(2020);
+        expect(dashboardContainerId).not.toBe(dashboardContainerId2);
+        await exec("docker", ["container", "stop", dashboardContainerId2]);
+      } else {
+        expect(dashboardContainerId).toBe("");
+      }
+    } catch (err) {
+      logger.error(err);
+    }
+  }, 30000);
```
evanlouie, line 73: 30 seconds?? does this test usually take that long?

src/commands/deployment/dashboard.ts
```diff
@@ -39,11 +39,42 @@ export const dashboardCommandDecorator = (command: commander.Command): void => {
...
+    "ancestor=samiyaakhtar/spektate:prod",
+    '--format="{{.ID}}"'
+  ]);
+  if (dockerOutput.length > 0) {
+    dockerOutput = dockerOutput.replace(/\n/g, " ");
+    dockerOutput = dockerOutput.replace(/"/g, "");
+    const containerIds = dockerOutput.split(" ");
+
+    let args = ["kill"];
+    args = args.concat(containerIds);
```
evanlouie: can be made a bit cleaner with : `const args = ["kill", ...containerIds]`

evanlouie (approved):

src/commands/deployment/dashboard.ts
```diff
@@ -39,11 +39,42 @@ export const dashboardCommandDecorator = (command: commander.Command): void => {
...
+ * Launches an instance of the spk dashboard
+ * @param port the port number to launch the dashboard
+ */
 export const launchDashboard = async (port: number): Promise<string> => {
   try {
     if (!(await validatePrereqs(["docker"], false))) {
       return "";
     }
+    // Clean previous dashboard containers
+    await cleanDashboarContainers();
```
samiyaakhtar: can we invoke this only if an argument such as `--kill-other-ports` (gotta think of a better one for this) or something is present? In case they are intentionally trying to launch it on different ports for a reason

samiyaakhtar: we can default it to true

src/commands/deployment/dashboard.test.ts
edaena: it took 18 seconds

samiyaakhtar: Yup it's been taking long on the azure builds on a bad day, but normally not this long for sure

edaena: Talked to samiya, 30 was a safe timeout number to account for delays from

src/commands/deployment/dashboard.ts
```diff
@@ -33,17 +38,51 @@ export const dashboardCommandDecorator = (command: commander.Command): void => {
...
+/**
+ * Cleans previously launched spk dashboard docker containers
+ */
+export const cleanDashboarContainers = async () => {
+  let dockerOutput = await exec("docker", [
+    "ps",
+    "-a",
+    "-q",
+    "--filter",
+    "ancestor=samiyaakhtar/spektate:prod",
```
samiyaakhtar: one more thing, can we take this `samiyaakhtar/spektate:prod`  from config? in case we change the name of the docker image in future, it's being referenced from `config.introspection!.dashboard!.image!`

edaena: good point!

samiyaakhtar (approved):

PR #101, opened 2019-10-24
Use a config param for private src repo access token

### LINKED ISSUE #707: [spk] Author field is showing empty for Azure DevOps repos in spk deployment tool
**Describe the bug:**
Dashboard shows empty cells under the "Author" column

**To Reproduce:**
Launch the dashboard for any setup that uses Azure DevOps repos

**Expected behavior:**
Author column should show the name of the author in Azure DevOps commits

**Screenshots:**
Screen Shot 2019-10-22 at 2 58 22 PM

**Additional context:**
The API call to get author in the spektate package may need some debugging.

samiyaakhtar:
Closes https://github.com/microsoft/bedrock/issues/707

When the source repo is private, spektate needs a personal access token to get Author information. 

Corresponding PR in spektate: https://github.com/microsoft/spektate/pull/42

### DIFF
--- a/spk-config.yaml
+++ b/spk-config.yaml
@@ -23,6 +23,7 @@ introspection:
     table_name: "storage-account-table-name" # Must be defined to run spk deployment commands
     partition_key: "storage-account-table-partition-key" # Must be defined to run spk deployment commands
     key: "storage-access-key" # Must be defined to run spk deployment commands. Use ${env:INTROSPECTION_STORAGE_ACCESS_KEY} and set it in .env file
+    source_repo_access_token: "source_repo_access_token" # Optional. Define this to see author names in the dashboard
 
     # Following 5 fields are needed only if using spk to onboard to use introspection tool
     service_principal_id: "service-principal-id"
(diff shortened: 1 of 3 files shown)


spk-config.yaml
```diff
@@ -23,6 +23,7 @@ introspection:
     table_name: "storage-account-table-name" # Must be defined to run spk deployment commands
     partition_key: "storage-account-table-partition-key" # Must be defined to run spk deployment commands
     key: "storage-access-key" # Must be defined to run spk deployment commands. Use ${env:INTROSPECTION_STORAGE_ACCESS_KEY} and set it in .env file
+    source_repo_access_token: "source_repo_access_token" # Optional. Define this to see author names in the dashboard
```
sarath-p: Is it different from `azure_devops.access_token? if it is, the source repo can belong to some other AzDo project.

samiyaakhtar: @sarath-p Yes it can be different if the src repo is private on GitHub and `azure_devops.access_token` is for the pipeline access token. Ran into this scenario yesterday with Tim's setup and he needed two separate tokens.

sarath-p: Do you expect the source repo access token work for hld and manifests repos too?  I am not sure if it requires the author name from hld and manifests repos as it progresses in the flow.

samiyaakhtar: @sarath-p We don't need the author information from HLD and manifests because only automated commits would go into it (ideally). So just a source repo access token is all we need

sarath-p (approved):

spk-config.yaml
edaena: Can you add in the comment that this is only required for private repos, and/or the other conditions under which this parameter is required?

edaena (approved):

spk-config.yaml
samiyaakhtar: Just addressed this

PR #102, opened 2019-10-24
Adding HLD lifecycle pipeline scaffold

bnookala:


evanlouie (approved):

src/lib/fileutils.ts
```diff
@@ -11,31 +11,55 @@ import {
...
+
+  const hldYaml = hldLifecyclePipelineYaml();
+  logger.info(`Writing hld-lifecycle.yaml file to ${azurePipelinesYamlPath}`);
+  fs.writeFileSync(azurePipelinesYamlPath, hldYaml, "utf8");
+};
+
 /**
  * Writes out the starter azure-pipelines.yaml file to `targetPath`
  *
  * @param targetPath Path to write the azure-pipelines.yaml file to
```
mtarng: Update documentation on params.

Also why would we need the filename to be variable?

mtarng: Followup, should we rename the pipelines to be more descriptive of what each one entails and should we also rename the functions for starterAzurePipelines to be reflect which pipeline it's for. @evanlouie @bnookala

mtarng (approved):

PR #111, opened 2019-10-31
Added command `spk ingress-route create`

### LINKED ISSUE #619: [SPK] Create IngressRoutes for new Services
**As a:**

SPK user

**I want:**

I want Ingress Routes to be automatically added to my services

**So that:**

I can have my tooling generate the necessary routing rules to accommodate rings

A method or a mechanism that _should_ add Ingress Routes to a service by: As
we're using Traefik ingress controller, we should create Traefik ingress routes.
This can be simply templatizing an ingress route. We want to create an Ingress
Route for the each of the rings defined in a bedrock.yaml file for my service,
and placing them in each subdirectory as a static manifest.

- [x] Produce a function/util inside spk that can output a materialized yaml
      output of a Traefik `IngressRoute` and configures it for a ring component
      in an HLD repository as a static manifest. This will be called by #717
  - `spk ingress-route create <service>  --ring <ring-name> --out /some/path/on/disk.yaml`
    - outputs the yaml file to `/some/path/on/disk.yaml`
  - `spk ingress-route create <service> --ring <ring-name>`
    - prints it out to console (useful for piping in a pipeline if needed)

**Additional context:**

In order for an application to be publicly accessible, it needs a routing rule
for the ingress controller (traefik, in this case)

Yes ~

evanlouie:
Added the `spk ingress-route create` command which allows users to create
Traefik `IngressRoute` manifests for the Traefik 2 IngressRoute CRD.

The command is setup such that a target `<service>` and `<port>` are required;
these will be used together to target a service in the Kubernetes cluster
matching `<service>:<port>`. If a `--ring` is provided, the target service
becomes `<service>-<ring>:<port>`.

The YAML output can either be outputted to a file with the `--output-file`
option, the option allows the user to specify any path-like directory on the
host; in the case of relative directories, it is resolved from the `cwd` of spk.
If a use leaves the `--output-file` blank or provides and empty string, the YAML
is printed to `stdout` with no additional log messages; this is useful if one
needs to pipe the output to in a pipeline.

closes https://github.com/microsoft/bedrock/issues/619

### DIFF
--- a/docs/ingress-route-management.md
+++ b/docs/ingress-route-management.md
@@ -1,0 +1,73 @@
+# Ingress Route Management
+
+Manage your Traefik `IngressRoute`s
+
+## Commands
+
+### create
+
+A command to create Kubernetes Traefik `IngressRoute`s.
+
+The primary usage of this command is to be used via pipelines to modifying your
+HLD with the appropriate manifests for ring based routing.
+
+```sh
+Usage: ingress-route create|c [options] <service> <port>
+
+Create a Traefik IngressRoute for a target <service>:<port>
+
+Options:
+  -r, --ring <ring>             the ring to deploy this service to, if provided the generated IngressRoute will target service `<service>-<ring>` (default: "")
+  --entry-point <entry-point>   the Traefik IngressRoute entryPoint; can be either 'web' or 'web-secure'; defaults to allowing all traffic if left blank (default: "")
+  --namespace <namespace>       a namespace to inject into the outputted Kubernetes manifest (default: "")
+  -o, --output-file <filepath>  filepath to output the IngressRoute YAML to; defaults to outputting to stdout (default: "")
+  -h, --help                    output usage information
+```
+
+#### Examples:
+
+**Outputting to stdout (useful if you need to pipe the output):**
+
+```sh
+$ spk ingress-route create my-service 80 --ring production --entry-point web --namespace my-fancy-ns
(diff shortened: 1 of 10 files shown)


mtarng (approved):

docs/ingress-route-management.md
```diff
@@ -0,0 +1,73 @@
...
+
+A command to create Kubernetes Traefik `IngressRoute`s.
+
+The primary usage of this command is to be used via pipelines to modifying your
+HLD with the appropriate manifests for ring based routing.
+
+```sh
+Usage: ingress-route create|c [options] <service> <port>
+
+Create a Traefik IngressRoute for a the target <service>:<port>
```
bnookala: "for a the" -> "for a"

docs/ingress-route-management.md
```diff
@@ -0,0 +1,73 @@
...
+          port: 80
+```
+
+**Outputting to to a file:**
+
+Note: the directory specified via `--output-file` will be automatically created
+if it does not exist.
+
+```sh
+$ spk ingress create my-service 80 --ring production --entry-point web --namespace my-fancy-ns --output-file ~/my/k8s/manifests/ingress.yaml
```
bnookala: should be `spk ingress-route create…`

bnookala (changes requested):

bnookala (approved):

PR #112, opened 2019-10-31
Adding reconcile command to spk

### LINKED ISSUE #717: [SPK] spk command to iterate over structured services and rings object and output fabrikate HLD
**As a:**

Developer

**I want:**

I want the option to modify my HLD repositories to add services or rings as needed via the `spk` cli tool.

**So that:**

I can create a HLD from a bedrock.yaml file, allowing me to add only the difference (new services) as a pull request to the HLD.

when I run `spk hld reconcile -b bedrock.yml -p /path/to/hld/repository/directory` it will:

- [ ] Verify that fabrikate is installed locally or in PATH. Link to download or install if it isn't.
- [ ] The logic flows for service with rings:
- For each service:
  - `mkdir service`
  - `fab add service`
  - `cd service`
  - `mkdir config`
  - for each ring:
    - `mkdir ring`
    - `fab add ring` - passing in the helm inputs encoded in the object structure from #716 
    - `cd ring`
    - `mkdir config`
    - `mkdir static`
    - `fab add static` (this will be where the ingressroute.yaml is stored)
    - call function to add traefik ingressroute manifest into the static directory via #619 

**Additional context:**

This shell script is meant to be invoked in a pipeline appliance to reconcile new services.
In a pipeline appliance, this shell script is to be invoked in an HLD repository, with the output being a difference against the existing HLD (or new services being added to the HLD). 

Considerations for reconciling the hld against bedrock.yaml, we should plan to support removing services and rings in the future.

Yes

bnookala:
- Adds reconcile command (for use in a pipeline)
- Reconcile only adds services to the HLD presently
- Reconcile adds ingress routes

Closes https://github.com/microsoft/bedrock/issues/717

### DIFF
--- a/src/commands/hld/reconcile.ts
+++ b/src/commands/hld/reconcile.ts
@@ -1,0 +1,182 @@
+import commander from "commander";
+import path from "path";
+import process from "process";
+import shelljs from "shelljs";
+
+import { Bedrock } from "../../config";
+import { logger } from "../../logger";
+
+import child_process from "child_process";
+import { writeFileSync } from "fs";
+import yaml from "js-yaml";
+import { promisify } from "util";
+
+import { TraefikIngressRoute } from "../../lib/traefik/ingress-route";
+
+const exec = promisify(child_process.exec);
+
+import { IBedrockFile } from "../../types";
+
+export const reconcileHldDecorator = (command: commander.Command): void => {
+  command
+    .command("reconcile <repository-name> <hld-path>")
+    .alias("r")
+    .description("Reconcile a HLD with the services tracked in bedrock.yaml.")
+    .action(async (repositoryName, hldPath) => {
+      try {
+        if (typeof repositoryName !== "string") {
+          throw new Error(
+            `repository-name must be of type 'string', ${typeof repositoryName} given`
+          );
+        }
+
+        if (typeof hldPath !== "string") {
+          throw new Error(
+            `hld-path must be of type 'string', ${typeof hldPath} given`
+          );
+        }
+
+        const fabrikateInstalled = shelljs.which("fab");
+
+        if (fabrikateInstalled === "") {
+          throw new Error(
+            `Error: Fabrikate not installed. Please fetch and install the latest version: https://github.com/microsoft/fabrikate/releases`
+          );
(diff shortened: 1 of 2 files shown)


src/commands/hld/reconcile.ts
```diff
@@ -0,0 +1,171 @@
...
+
+import { IBedrockFile } from "../../types";
+
+export const reconcileHldDecorator = (command: commander.Command): void => {
+  command
+    .command("reconcile")
+    .alias("r")
+    .description("Reconcile a HLD with the services tracked in bedrock.yaml.")
+    .option(
+      "-, --hld-path <hld-path>",
```
mtarng: no short hand option/make sure it doesn't override -h.

src/commands/hld/reconcile.ts
```diff
@@ -0,0 +1,182 @@
...
+          );
+        }
+
+        if (typeof hldPath !== "string") {
+          throw new Error(
+            `hld-path must be of type 'string', ${typeof hldPath} given`
+          );
+        }
+
+        const fabrikateInstalled = shelljs.which("fab");
```
evanlouie, line 39: for future reference: https://github.com/microsoft/fabrikate/issues/243

evanlouie (approved):

src/commands/hld/reconcile.ts
bnookala: 👍

mtarng (approved):

src/commands/hld/reconcile.ts
```diff
@@ -0,0 +1,182 @@
...
+};
+
+/**
+ * Runs a command via `exec` and captures the results, logging the command
+ * that will be run, and the results of that command.
+ *
+ * @param commandToRun String version of the command that must be run
+ * @throws An Error containing the logs captured from stderr
+ */
+const execAndLog = async (commandToRun: string) => {
```
mtarng, line 171: Perhaps we'll want this in a lib later. A bit confusing with this reference and us having an `exec` library too.

bnookala: fair

PR #117, opened 2019-11-01
Nate.infra.generate: Support for source validation and cloning

NathanielRose:
Completes Task 1 of 3 in issue: [653](https://github.com/microsoft/bedrock/issues/653)
-  Check if the source repo to a temp directory is provided in definition.json, if not validate a remote exists and clone the repo.


andrebriggs:
@NathanielRose is there a reason we including Terragrunt artifacts?

src/commands/infra/generate.test.ts
```diff
@@ -0,0 +1,64 @@
...
+} from "../../logger";
+import {
+  validateDefinition,
+  validateRemoteSource,
+  validateTemplateSource
+} from "./generate";
+
+beforeAll(() => {
+  enableVerboseLogging();
+  jest.setTimeout(10000);
```
yradsmikham, line 17: Is there a reason for setting a specific timeout?

src/commands/infra/generate.test.ts
```diff
@@ -0,0 +1,64 @@
...
+});
+
+afterAll(() => {
+  disableVerboseLogging();
+  jest.setTimeout(5000);
```
yradsmikham, line 22: Same question as above.

src/commands/infra/mocks/definition.json
```diff
@@ -0,0 +1,42 @@
...
+    "subnet_prefixes": "<insert value>",
+    "vnet_name": "<insert value>",
+    "subnet_name": "<insert value>",
+    "network_plugin": "azure",
+    "network_policy": "azure",
+    "oms_agent_enabled": "false",
+    "enable_acr": "false",
+    "acr_name": "<insert value>"
+  }
+}
```
yradsmikham, line 42: Committing a "mock" definition.json file to master branch for the sake of unit testing might not be ideal, especially if definition.json files could vary. Something we have done in `spk infra scaffold` is create a JSON object as part of the unit test (instead of relying on a file), and testing on components from the JSON object. Thoughts?

yradsmikham: Refer to this: https://github.com/CatalystCode/spk/blob/baf4adb74317295c1e057b965ba49df3834a82a8/src/commands/infra/scaffold.test.ts#L18

src/commands/mocks/azure-simple/terragrunt.hcl
```diff
@@ -0,0 +1,28 @@
...
+    vnet_name = "" 
+    service_cidr = "10.0.0.0/16" 
+    dns_ip = "10.0.0.10" 
+    docker_cidr = "172.17.0.1/16" 
+    address_space = "10.10.0.0/16" 
+    subnet_prefix = "10.10.1.0/24" 
+    network_plugin = "azure" 
+    network_policy = "azure" 
+    oms_agent_enabled = "false"
+  }
```
yradsmikham: Let's remove this. This is carryover from another branch or PR.

src/commands/infra/generate.ts
```diff
@@ -0,0 +1,190 @@
...
+    const definitionJSON = JSON.parse(data);
+    // TO DO : Check for malformed JSON
+    if (!(definitionJSON.template && definitionJSON.source)) {
+      logger.info(
+        `The definition.json file is invalid. There is a missing field for the definition file's sources. Template: ${definitionJSON.template} source: ${definitionJSON.source} version: ${definitionJSON.version}`
+      );
+      return [];
+    }
+    logger.info(
+      `Checking for locally stored template: ${definitionJSON.template} from remote repoitory: ${definitionJSON.source} at version: ${definitionJSON.version}`
```
yradsmikham: misspelled "repository"

yradsmikham (changes requested):

yradsmikham:
@NathanielRose Let's also update the documentation on how to use `spk infra generate`

src/commands/infra/generate.test.ts
NathanielRose: Since in my unit test, I actually clone a repo, It requires an increase in timeout by 5 seconds

src/commands/infra/mocks/definition.json
NathanielRose: The unit test checks the presence of a definition.json in the provided project folder. From there it checks the source url for a repository. I can see the definition.json becoming more complex for future unit tests we want to use.

src/commands/mocks/azure-simple/terragrunt.hcl
NathanielRose: Removed terragrunt.hcl @andrebriggs

src/commands/infra/generate.ts
NathanielRose: Modified

NathanielRose:
> @NathanielRose Let's also update the documentation on how to use `spk infra generate`

Updated the docs, but only stated current functionality, source caching.

yradsmikham (approved):

docs/cloud-infra-management.md
```diff
@@ -129,3 +130,19 @@ definition.json
...
+containing a `definition.json` that contains a `source`, `template` and
+`version` to obtain the terraform template files.
+
+It will do the following (**In Progress**):
+
+- Check if a provided project folder contains a `definition.json`
+- Check if the terraform template `source` provided has a valid remote
+  repository
+- Cache the master version of the provided `source` repository locally in
+  `~/.spk/templates`
```
jmspring, line 148: Is there ever a case where SPK config might be somewhere other than ~/.spk?  Also - how are templates differentiated?

src/commands/infra/generate.ts
```diff
@@ -0,0 +1,190 @@
+import child_process from "child_process";
+import { getMaxListeners } from "cluster";
+import commander from "commander";
+import fs, { chmod } from "fs";
+import * as os from "os";
+import path from "path";
+import simpleGit from "simple-git/promise";
+import { logger } from "../../logger";
+
+const spkTemplatesPath = os.homedir() + "/.spk/templates";
```
jmspring, line 10: Will there ever be a use case where this will need to differ?  Can it be recreated with an existing cluster?

src/commands/infra/generate.ts
```diff
@@ -0,0 +1,190 @@
...
+/**
+ * Checks if definition.json is present locally to provided project path
+ *
+ * @param projectPath Path to the definition.json file
+ */
+export const validateDefinition = async (
+  projectPath: string
+): Promise<boolean> => {
+  try {
+    // If templates folder does not exist, create cache templates directory
```
jmspring, line 60: Why does this need to be created here?  Seems to have no impact on the functionality itself.

src/commands/infra/generate.ts
```diff
@@ -0,0 +1,190 @@
...
+export const validateRemoteSource = async (
+  definitionJSON: string[]
+): Promise<boolean> => {
+  try {
+    const [source, template, version] = definitionJSON;
+    // Converting source name to storable folder name
+    const httpReg = /^(.*?)\.com/;
+    const punctuationReg = /[^\w\s]/g;
+    let sourceFolder = source.replace(httpReg, "");
+    sourceFolder = sourceFolder.replace(punctuationReg, "_").toLowerCase();
```
jmspring: Maybe document what the expected directory structure/name will look like?

src/commands/infra/generate.ts
```diff
@@ -0,0 +1,190 @@
...
+    const sourcePath = path.join(spkTemplatesPath, sourceFolder);
```
jmspring: Should the presence of spkTemplatesPath be checked for here rather than in the prior function?

src/commands/infra/generate.ts
```diff
@@ -0,0 +1,190 @@
...
+        `Provided source template folder was not found, attempting to clone the template source repo locally.`
+      );
+      fs.mkdirSync(sourcePath);
+    } else {
+      logger.info(
+        `Source template folder found. Checking remote existence of remote repository`
+      );
+    }
+    // Checking for git remote
+    require("simple-git")(sourcePath).listRemote(
```
jmspring: The require is done once for "simpleGit" which is instantiated with "git = simpleGit()", why the use of require here?  Is it possible to not do the require multiple times?

src/commands/infra/generate.ts
```diff
@@ -0,0 +1,190 @@
...
+          if (!result) {
+            logger.error(
+              `Unable to clone the source remote repository. Does the remote repo exist? Do you have the rights to access it? ${result}`
+            );
+            return false;
+          } else {
+            logger.info(
+              `Checking if source repo: ${source} has been already cloned to: ${sourcePath}.`
+            );
+            require("simple-git")(sourcePath).revparse(
```
jmspring: As above re: using require

jmspring (commented):
Some questions / notes.  Pending answers, LGTM

docs/cloud-infra-management.md
NathanielRose: This may be a customization demand that we can support. This needs to be raised to the larger team though if we want to have this as a path provided through `spk init`. For simplicity, we can use the `/.spk` as this is similar to what az cli does.

src/commands/infra/generate.ts
NathanielRose: Related to comment before around the cache folder. We can support this function, just needs to be built into `spk init`

src/commands/infra/generate.ts
NathanielRose: I had this originally in the decorator but it needs to be inside an async function I believe. We alleviate the user having to even care about maintaining the cache template folder so generate creates it for them. I can also see the entire `/.spk` cache folder as something teams can use t share cli environments maybe. Thoughts?

src/commands/infra/generate.ts
NathanielRose: Updated.

src/commands/infra/generate.ts
NathanielRose: `fs` during testing was not waiting for the return of checking for the folder and bypassing the check for the template. I placed it into another function to force the promise to await in `fs`. Open to suggestions

src/commands/infra/generate.ts
NathanielRose: Simple-git typescript definitions are constructed weird where the `sourcePath` maps to where I want the git command to run. The alternative would be to change the working path every time to the respective folder `process.chdir(${sourcePath)` It would still have redundancy. Open to suggestions.

src/commands/infra/generate.ts
NathanielRose: Responded.

docs/cloud-infra-management.md
jmspring: After talking to Saiya and Andre, the expectation is .spk for now, so my questions around there can be addressed in a "feature enhancement" issue that would be low pri.

src/commands/infra/generate.ts
jmspring: K

src/commands/infra/generate.ts
jmspring: Keep as is.

src/commands/infra/generate.ts
jmspring: Resolved per above.

jmspring (approved):
LGTM

src/commands/infra/generate.ts
evanlouie: does having `import simpleGit from "simple-git"` on top doing `simpleGit(sourcePath)` not work?

yradsmikham:
It appears that this PR does not handle repo tags for versioning. Is this correct?

NathanielRose:
> It appears that this PR does not handle repo tags for versioning. Is this correct?

Correct as mentioned in [653](https://github.com/microsoft/bedrock/issues/653)

PR #131, opened 2019-11-13
Onboard refactoring

### LINKED ISSUE #681: Refactor service introspection `onboard` command to make Key Vault optional
**As a:**
Developer/Operator

**I want:**

- [x] Configure a single storage account for service introspection data and use access keys stored from Azure Key vault with a fall back option to spk config. Currently the `onboard` sets the storage key in keyvault secret, but not used while loading configuration object for `spk deployment/dashboard`.
- [x] Make --key-vault parameter optional in `onboard` so that storage access key can be configured manually in spk config file when customers are not using the Key Vault in their environment.
- [x] Create a table within the storage account when it does not exist in whether a newly created storage account or an existing storage account in `onboard` command
- [x] Return storage account output in json format from `onboard` command to help with automation. For ex: the access key in the output can be used to create a variable in Variable Group subsequently.

**So that:**

Use storage account access keys from either spk config or Azure Key Vault. 
Automate Service Introspection setup steps.

**Additional context:**

sarath-p:
Closes microsoft/bedrock#681.

In addition to closing above, the following tasks are addressed.

- Fixed storing environment/shell variable values in ~./spk/config.yaml.
- Saving storage account and table names in config.yaml without a manual step
- Updated test cases which made some mock files obsolete


src/config.ts
```diff
@@ -180,3 +219,34 @@ export const loadConfiguration = (filepath: string = defaultFileLocation()) => {
...
+export const writeConfigToDefaultLocation = async (filePath: string) => {
+  try {
+    const data = yaml.safeDump(readYaml<IConfigYaml>(filePath));
+    await saveConfig(data);
+  } catch (err) {
+    logger.error(
+      `Error occurred while writing config to default location. \n ${err}`
+    );
+  }
+};
```
samiyaakhtar, line 253: Ideally, would be nice to do this in a separate PR since it's not related to onboard

docs/service-introspection.md
```diff
@@ -110,7 +110,10 @@ Options:
 
 Onboard to use the service introspection tool. This will create a storage
 account if it does not already exist in your subscription in the given
-`resource-group`.
+`resource-group`. The storage table will also be created in a newly created or
+in an existing storage account. When Azure Key Vault argument is specified, a
+secret with Azure storage access key will be cerated otherwise, the storage
```
andrebriggs: Typo at `cerated`

docs/service-introspection.md
```diff
@@ -126,11 +129,16 @@ Usage:
...
 
 Options:
-  -n, --storage-account-name <storage-account-name>                Account name for the storage table
-  -l, --storage-location <storage-location>                        Azure location for Storage account and resource group when they do not exist
-  -r, --storage-resource-group-name <storage-resource-group-name>  Name of the resource group for the storage account
-  -k, --key-vault-name <key-vault-name>                            Name of the Azure key vault
-  -h, --help                                                       Usage information
+  -s, --storage-account-name <storage-account-name>                 Azure storage account name
+  -t, --storage-table-name <storage-table-name>"                    Azure storage table name
+  -l, --storage-location <storage-location>                         Azure location to create new storage account when it does not exist
```
andrebriggs, line 140: Are we not offering specifying this in the config file? I don't see `storage-location` in the config file or `types.d`

src/commands/deployment/get.ts
```diff
@@ -151,8 +153,9 @@ export const getDeployments = (
 /**
  * Initializes the pipelines assuming that the configuration has been loaded
  */
-const initialize = () => {
+const initialize = async () => {
   const config = Config();
+  const key = await Config().introspection!.azure!.key;
```
andrebriggs, line 158: Out of curiosity why do we specifically need to `await` on this field as opposed to the others?

src/commands/deployment/onboard.ts
```diff
@@ -18,82 +30,244 @@ export const onboardCommandDecorator = (command: commander.Command): void => {
...
+
+  const isExist = await isStorageAccountExist(resourceGroup, accountName, opts);
 
-  if (key === undefined) {
-    const errorMessage: string = `Storage account ${accountName} access keys in resource group ${resourceGroup}is not available`;
-    logger.error(errorMessage);
-    throw new Error(errorMessage);
+  // Storage account does not exist so create it.
+  if (isExist === false) {
+    if (location === undefined || location === null || location === "") {
```
andrebriggs: We probably have many over instances in `spk` where we do this sort of checking. Is there a chance to refactor in to a "null or empty" string check method?

src/commands/deployment/onboard.ts
```diff
@@ -18,82 +30,244 @@ export const onboardCommandDecorator = (command: commander.Command): void => {
...
+
+/**
+ * Checks arguments for undefined or null and returns errors
+ *
+ * @param storageAccountName The Azure storage account name
+ * @param storageTableName The Azure storage table name
+ * @param storageResourceGroup Name of Azure reesource group
+ * @param accessOpts The Azure subscription access options
+ */
+export const validateRequiredArguments = async (
```
andrebriggs, line 256: This is another area where we should used the same argument verification method. Perhaps we pass in an array of expected args that way others can reuse this logic?

src/lib/azure/keyvault.test.ts
```diff
@@ -0,0 +1,60 @@
+import uuid from "uuid/v4";
```
andrebriggs, line 1: Thanks for creating tests for components like these 👍🏾 For the sake of reviewers it might be good in the future to create a smaller PRs that have unit test and implementation for components like this. That way each PR has a clear focus

src/lib/azure/storage.test.ts
```diff
@@ -18,28 +27,63 @@ afterAll(() => {
...
+    expect(account).toBeUndefined();
+  });
 });
 
 describe("get storage account key", () => {
-  it("should create storage account", async () => {
-    logger.info(`called mock get key`);
-    const key = await getStorageAccountKey("epi-test", "testsarathp3");
-    expect(key).toBe("mock access key");
+  test("should fail getting storage account key when arguments are not specified", async () => {
```
andrebriggs, line 70: Good job creating negative error cases

andrebriggs (changes requested):

src/config.ts
sarath-p: Agreed. I was trying to reuse the method to save storage and table names in the config, but we can  do this in a separate PR.

samiyaakhtar: @sarath-p No this is fine for now! I understand the PR a lot more after testing it 👍

docs/service-introspection.md
sarath-p: Good catch.

docs/service-introspection.md
sarath-p: No, we are not offering the location in the `config` as well as `types.d.ts` since it is only needed once while onboarding the introspection which creates the storage account if necessary. Let me know if I miss any scenarios that requires it to be in config so we can add it.

src/commands/deployment/get.ts
sarath-p: The `key` property implements a lazy `getter` that fetches the storage access key either from `Azure KeyVault` or `Config` file. It uses a `Promise` to avoid circular calls i.e. while loading `Config()` object first time, it makes a call to `keyVault` which requires `service_principal_id`, `service_principal_secret`, and `tenant_id` from the `Config()` object that goes into a circular loop. The `Promise` allows us to delay fetching the key value until any part of the code requires it. Does it makes sense?

src/commands/deployment/onboard.ts
sarath-p: We can create a reusable method, but `typescript` offers a simpler syntax like `if (!location){` which we are using in some places. I changed this instance in this PR and we can create a refactoring task to make it consistent.

src/commands/deployment/onboard.ts
sarath-p: Good idea, We can create a separate issue to implement a method and refactor all commands arguments.

src/lib/azure/keyvault.test.ts
sarath-p: Agreed. Added a validation for input arguments and thought unit tests are needed to test it.

samiyaakhtar (approved):

src/commands/deployment/onboard.ts
```diff
@@ -18,82 +26,245 @@ export const onboardCommandDecorator = (command: commander.Command): void => {
...
   );
 
-  const key = await getStorageAccountKey(resourceGroup, accountName);
+  let storageAccount: StorageAccount | undefined;
+
+  const isExist = await isStorageAccountExist(resourceGroup, accountName, opts);
+
+  // Storage account does not exist so create it.
+  if (isExist === false) {
+    // if (location === undefined || location === null || location === "") {
```
edaena: remove comment

spk-config.yaml
```diff
@@ -30,3 +30,4 @@ introspection:
     subscription_id: "subscription-id"
     tenant_id: "tenant-id"
     resource_group: "resource-group-name"
+key_vault_name: # azure key vault name"
```
edaena, line 34: do we indicate that this is optional? should it be commented out?

sarath-p: The Key Vault is optional but other parts of `spk` might make it required so I updated introspection docs to inform the user that Key Vault is optional.

src/commands/deployment/onboard.ts
sarath-p: Good catch, removed it!

edaena (approved):

src/commands/deployment/onboard.test.ts
```diff
@@ -159,22 +170,122 @@ describe("validateRequiredArguments", () => {
...
+        introspection: {
+          azure: {
+            account_name: undefined,
+            table_name: undefined
+          }
+        }
+      };
+
+      // create config file in test location
+      fs.writeFileSync(testConfigFile, yaml.safeDump(data));
```
andrebriggs, line 213: @sarath-p why do we need the config file?

andrebriggs (approved):

src/commands/deployment/onboard.test.ts
sarath-p: The test scenario is to check whether `onboard` can update an existing config file with storage account and table names are undefined/null and this particular line is storing the config file with empty values before updating.

src/commands/deployment/onboard.ts
```diff
@@ -18,82 +25,297 @@ export const onboardCommandDecorator = (command: commander.Command): void => {
...
+    accountName,
+    tableName,
+    accessKey!
+  );
+
+  logger.debug(`tabled created: ${tableCreated}`);
+
+  // print newly created storage account
+  if (storageAccount !== undefined) {
+    logger.info(`${JSON.stringify(storageAccount, null, 2)}\n`);
```
evanlouie: don't need trailing `\n`

src/commands/deployment/onboard.ts
```diff
@@ -18,82 +25,297 @@ export const onboardCommandDecorator = (command: commander.Command): void => {
...
+  }
+
+  if (storageAccount !== undefined && tableCreated === true) {
+    logger.info(
+      `Storage account ${accountName} and table ${tableName} are created.\n`
```
evanlouie: don't need trailing `\n`

src/commands/deployment/onboard.ts
```diff
@@ -18,82 +25,297 @@ export const onboardCommandDecorator = (command: commander.Command): void => {
...
+    );
+  } else if (storageAccount === undefined && tableCreated === true) {
+    logger.info(
+      `Table ${tableName} is created in existing storage account ${accountName}.\n`
```
evanlouie: don't need trailing `\n`

src/commands/deployment/onboard.ts
```diff
@@ -18,82 +25,297 @@ export const onboardCommandDecorator = (command: commander.Command): void => {
...
+    );
+  } else if (storageAccount === undefined && tableCreated === false) {
+    logger.info(
+      `Both storage account ${accountName} and table ${tableName} exist.\n`
```
evanlouie: don't need trailing `\n`

src/commands/deployment/onboard.ts
```diff
@@ -18,82 +25,297 @@ export const onboardCommandDecorator = (command: commander.Command): void => {
...
+    );
+
+    await setSecret(keyVaultName, `${accountName}Key`, accessKey!, opts);
+  } else {
+    // notify the user to set the environment variable with storage access key
+    logger.info(
+      `Please set the storage account access key in environment variable INTROSPECTION_STORAGE_ACCESS_KEY before issuing any deployment commands.\n`
+    );
+
+    logger.info(`Storage account ${accountName} access key: ${accessKey}\n`);
```
evanlouie: don't need trailing `\n`

src/commands/deployment/onboard.ts
```diff
@@ -18,82 +25,297 @@ export const onboardCommandDecorator = (command: commander.Command): void => {
...
+  if (keyVaultName) {
+    logger.debug(
+      `Calling setSecret with storage account primary key ${accessKey} and ${keyVaultName}`
+    );
+
+    await setSecret(keyVaultName, `${accountName}Key`, accessKey!, opts);
+  } else {
+    // notify the user to set the environment variable with storage access key
+    logger.info(
+      `Please set the storage account access key in environment variable INTROSPECTION_STORAGE_ACCESS_KEY before issuing any deployment commands.\n`
```
evanlouie: don't need trailing `\n`

evanlouie (approved):

PR #132, opened 2019-11-13
spk infra generate (iteration 2)

### LINKED ISSUE #721: Generate multi node spk template
**As a:**
Developer

**I want:**
to generate Terraform files for a multi-cluster based on parent-leaf definitions

**So that:**
I can avoid the need to manually update individual cluster configurations when a shared variable has changed.

This is an extension to https://github.com/microsoft/bedrock/issues/749.

Upon generation, `spk infra generate` (e.g. `spk infra generate east`) should:
- [x] Parse `definition.json` from parent directory (current working directory)
- [x] Replicate a hierarchy similar to the target directory. For example:

```
discovery-service
    |- definition.json
    |- east
        |- definition.json
    |- central
        |- definition.json

discovery-service-generated
    |- east
        |- main.tf
        |- variables.tf
        |- spk.tfvars
    |- central
        |- main.tf
        |- variables.tf
        |- spk.tfvars
```
- [x] Iterate through definition.json files from parent and leaf directories
- [x] Write to spk.tfvars for cluster environment

Yes

yradsmikham:
- Recursively reads `definition.json` files from parent to leaf directories
- Creates a `generated` hierarchy resembling definition hierarchy
- Copies appropriate Terraform templates to `generated` directory

Closes https://github.com/microsoft/bedrock/issues/721.


src/commands/infra/generate.test.ts
```diff
@@ -83,10 +84,8 @@ describe("Validate template path from a definition.json", () => {
 describe("Validate spk.tfvars file", () => {
   test("Validating that a spk.tfvars is generated and has appropriate format", async () => {
     const mockProjectPath = "src/commands/infra/mocks";
-    const generateTfvars = await generateSpkTfvars(
-      mockProjectPath,
-      mockProjectPath
-    );
+    const definitionJSON = await readDefinitionJson(mockProjectPath);
```
andrebriggs, line 87: @yradsmikham this unit test appears to be doing file system operations. If we separated the filesystem input portion from the actual JSON objects in `writeSpkTfvars` it seems we could have unit tests that are not dependent on looking at the file system. This makes the tests more portable as well. 

We should consider decoupling the filesystem operations from the business logic in `writeSpkTfvars` when making further PRs

yradsmikham: The challenge is decoupling `fs` operations in `writeSpkTfvars` because this is the one method in the command that will create and write to the `spk.tfvars`. So, we _can_ decouple_ fs operations into another method, and call the method within `writeSpkTfcars`, but it would only make the script more complex, and lead us to write another unit test to test additional methods.

Do you think this is more of an integration test? If so, we could move this over.

andrebriggs: @yradsmikham I would argue it makes the business logic easier to see and less complicated since your method is smaller. I can share an example offline with you about this when I get time. All I'm saying is for future PRs let's think about it. 

Technically this is an integration test. I'm saying we should avoid disk and network I/O in unit tests (for speed, portability, and designing our code to have clear dependencies). If we can work towards that great but I understand sometimes it may be difficult or non-obvious. 

One thing I realized from this is that we don't publish results of how long each "unit" test is taking. Once we have those result we can make a data driven decision on if this is worth the time.

src/commands/infra/generate.ts
```diff
@@ -173,14 +170,78 @@ export const validateRemoteSource = async (
...
+ */
+export const generateConfig = async (projectPath: string): Promise<void> => {
+  try {
+    // First, search for definition.json in current working directory
+    const templatePath = await parseDefinitionJson(projectPath);
+    const files = fs.readdirSync(".");
+    const cwdPath = process.cwd();
+    files.forEach(async file => {
+      // If there exists a definition.json, then read file
+      if (file === "definition.json") {
```
andrebriggs: @yradsmikham another example of the point I'm raising is that in this loop we are wholly dependent on there being a file system and a collection of files to iterate. When we find a file we read it via ` readDefinitionJson` and get an object back (`parentDefinitionJSON` or `leafDefinitionJSON`). From a unit test perspective we care about what is inside the defintionJSON _objects_.  

The meat of this loop is starts on line 194. That loop is our business logic that we can test without disk/network I/O.  Imagine if we had another method that started at line 194 and took the `parentDefinitionJSON` and `leafDefinitionJSON` objects  as method parameters and returned `parentDefinitionJSON.variables`

Just explaining the thought process. No changes needed

andrebriggs (approved):

NathanielRose:
@yradsmikham So I think we may need more guidance in the documentation on how to use generate and what the expected functionality would be. (Discretion: It could just be me not following directions :smile: )

Here is my Tree structure after scaffolding:
```
discovery-service
    |- definition.json
    |- east
    |   |- definition.json
    |- west
        |- definition.json
```

Currently, I am seeing 2 different behaviors:

- With the tree structure above I run `spk infra generate --project discovery-service` and I generate an improperly named folder `-generate` with only 1 level of deployment templates. See below:

```
naterose@naterose-Blade:~/Desktop/Microsoft/fy20/spk-infra/spk/dist/discovery-service$ ls
definition.json  east  -generated  west
naterose@naterose-Blade:~/Desktop/Microsoft/fy20/spk-infra/spk/dist/discovery-service/generated$ ls
backend.tfvars  main.tf  README.md  spk.tfvars  terraform.tfvars  variables.tf
```

- Inside the `discovery-service` folder, I run `spk infra generate` which creates this new tree structure below:

```
discovery-service-generated
    |- backend.tfvars
    |- main.tf
    |- readme.md
    |- spk.tfvars
    |- terraform.tfvars
    |- variables.tf
    |-home
        |- /naterose/Desktop/Microsoft/fy20/spk-infra/spk/dist/discovery-service 
```
There's a home folder that contains all the folders in the path to discovery-service but it still is missing `east` and `west` generated folder.

> Also, When we run `terraform plan` later in the pipeline, I believe `terraform.tfvars` may have precedence over `spk.tfvars`. Should we add the `.backup` to it to avoid conflict?

src/commands/infra/generate.ts
yradsmikham: Gotcha, that's a good point. I will make note of that and have it be an improvement for the next iteration. Thanks!

yradsmikham:
@NathanielRose apologies, if the documentation was not well-written. I'll spend more time improving it. Although, I'm trying to reproduce the issue you're experiencing, and am unable to do so.

So, there is such thing as running `spk infra generate` in a specific working directory in order to produce a specific outcome, and I'm thinking you could be running the command in different working directory. With your current tree structure, if you were to run `spk infra generate -p discovery-service` a level _above_ the `discovery-service` directory, then this will produce a `discovery-service-generated` directory alongside the `discovery-service` level. This is considered a "single-tree" generation. So you would have:

```
discovery-service
    |- definition.json
    |- east/
    |   |- definition.json
    |- west/
        |- definition.json
discovery-service-generated
    |- backend.tfvars
    |- main.tf
    |- readme.md
    |- spk.tfvars
    |- terraform.tfvars
    |- variables.tf
```

If you want to generate for a specific _region_ of the `discovery-service`, you should first `cd` into the `discovery-service` directory, and then run `spk infra generate -p east`. This will create the following:

```
discovery-service
    |- definition.json
    |- east/
    |   |- definition.json
    |- west/
        |- definition.json
discovery-service-generated
    |- east/
        |- main.tf
        |- readme.md
        |- spk.tfvars
        |- terraform.tfvars
        |- variables.tf
```

I will go ahead and update the documentation, so it has better instructions on how to use `spk infra generate`. I will also re-enable appending `.backup` to `terraform.tfvars`.

NathanielRose:
@yradsmikham Here is an exact repro:
- Scenario 1
``` sh
$ ./spk-linux infra scaffold --name discovery-service --source https://github.com/microsoft/bedrock --version "v0.12.0" --template cluster/environments/azure-single-keyvault
$ ./spk-linux infra scaffold --name discovery-service/east --source https://github.com/microsoft/bedrock --version "v0.12.0" --template cluster/environments/azure-simple
$ ./spk-linux infra scaffold --name discovery-service/west --source https://github.com/microsoft/bedrock --version "v0.12.0" --template cluster/environments/azure-simple
$ ./spk-linux infra generate --project discovery-service/
$ cd discovery-service/
$ ls
definition.json  east  -generated  west
$ Using File Explorer to rename -generated to discovery-generated
$ cd discovery-generated/
$ ls
backend.tfvars  main.tf  README.md  spk.tfvars  terraform.tfvars  variables.tf
```

- Scenario 2
``` sh
$ ./spk-linux infra scaffold --name discovery-service --source https://github.com/microsoft/bedrock --version "v0.12.0" --template cluster/environments/azure-single-keyvault
$ ./spk-linux infra scaffold --name discovery-service/east --source https://github.com/microsoft/bedrock --version "v0.12.0" --template cluster/environments/azure-simple
$ ./spk-linux infra scaffold --name discovery-service/west --source https://github.com/microsoft/bedrock --version "v0.12.0" --template cluster/environments/azure-simple
$ cd discovery-service/
$ ./spk-linux infra generate
$ cd ../discovery-service-generated/
$ ls
backend.tfvars  home  main.tf  README.md  spk.tfvars  terraform.tfvars  variables.tf
```
Where home has the folder path to discovery-service/

NathanielRose (approved):
lgtm - Works on Linux/windows
The current behavior for `generate` is based on - https://github.com/CatalystCode/bedrock-end-to-end-dx#generating-deployable-cluster-terraform-scripts
Enhanced Future functionality depicted here  - https://github.com/microsoft/bedrock/issues/791

src/commands/infra/generate.test.ts
jmspring: Second the removal of fs ops from unit tests.

src/commands/infra/scaffold.ts
```diff
@@ -135,7 +135,7 @@ export const renameTfvars = async (dir: string): Promise<void> => {
   try {
     const tfFiles = fs.readdirSync(dir);
     tfFiles.forEach(file => {
-      if (file.substr(file.lastIndexOf(".") + 1) === "tfvars") {
+      if (file === "terraform.tfvars") {
```
jmspring, line 138: I would still prefer that we not copy the "tfvars" file in the first place.  If there are multiple tfvars files, the above could cause issues.

jmspring (dismissed):
The renaming of "terraform.tfvars" loop only renames the one file, not the others.  Further, it continues to iterate once that single file is renamed.  See note for preferred fix and possible error.

src/commands/infra/generate.ts
yradsmikham: I managed to resolve this in my latest commit!

src/commands/infra/scaffold.ts
yradsmikham: resolved!

src/commands/infra/generate.test.ts
yradsmikham: fixed in my latest commit!

yradsmikham:
@jmspring I removed the renaming functionality, and refactored the `copyTfTemplate` functionality to exclude the `terraform.tfvars` file.

src/commands/infra/generate.ts
```diff
@@ -224,41 +305,43 @@ export const parseDefinitionJson = async (projectPath: string) => {
...
     });
+    return tfVars;
   } catch (err) {
-    logger.error(err);
+    logger.error(`There was an error with generating the spk tfvars object.`);
+    return err;
   }
 };
 
+export const writeToSpkTfvarsFile = async (
```
edaena, line 342: Missing documentation

edaena, line 342: Missing function documentation

samiyaakhtar (approved):

PR #137, opened 2019-11-14
Service Introspection self-test

edaena:
[Bedrock self-test to validate service introspection](https://github.com/microsoft/bedrock/issues/729)


src/commands/deployment/self-test.ts
```diff
@@ -0,0 +1,177 @@
+import commander from "commander";
```
edaena: pending self-test.test.ts

src/commands/deployment/self-test.ts
```diff
@@ -0,0 +1,177 @@
...
+  ).then(async entries => {
+    let entryToDelete: any;
+    logger.info("Deleting test data...");
+    let foundEntry = false;
+
+    try {
+      for (const entry of entries) {
+        entryToDelete = entry;
+
+        if (entry.p1 && entry.p1._ === buildId) {
```
samiyaakhtar: nit: I'm thinking if we should use something other than a buildId for this verification (`writeSelfTestData` should perhaps return the RowKey instead of the `buildId`) in case there's two entries in the table with this build Id?

samiyaakhtar:
Looks good so far @edaena

src/commands/deployment/self-test.ts
edaena: Do you think that's needed even though this is already only fetching the entries for the `self-test` service?

edaena: @samiyaakhtar also the buildId is random

samiyaakhtar: @edaena oh yeah then it should be fine

samiyaakhtar (approved):

sarath-p (approved):
Looks good!

PR #138, opened 2019-11-14
Removed Service Creation Options from project init

### LINKED ISSUE #719: [SPK] Refactor `spk project` to have the same functionality with both mono and non-mono repositories
**As a:**
spk user
**I want:**
spk project init to work for both mono and non mono repositories
**So that:**
the onboarding experience is the same for both
`spk project init` should create the base files needed for a bedrock project: `bedrock.yaml`, `maintainers.yaml`, `.gitignore` and anything else that's share for mono-repo and single application repositories. It should NOT write out the `azure-pipeline.yaml` file, as it does not know where the service is located, or if a user will onboard more later.

A Follow up task would be to give `spk service create` a recursive or directory flag to onboard multiple services after running `spk project init`. Alternatively, to be explicit, a user must manually add each service.

- [ ] `spk project init` no longer takes the -mono repo flag nor the -p packages directory flags. spk project init will still create the hld-pipeline.yaml but will not process any services.
- [ ] Update documentation to reflect these changes and that all service onboarding will be done manually though `spk service create`

**Additional context:**

Yes

evanlouie:
- Removed the `--mono-repo` (`-m`) and `--packages-dir` (`-d`) option from
  `project init`, removing the service onboarding flow from the `project init`
  flow.
- Removed the `--variable-group-name` from `project init`, removing the ability
  to inject a default variable group name into a service's azure-pipelines.yaml.
- Add and removed some test for `project/init.test.ts` to ensure the removed
  flows don't break anything.

closes https://github.com/microsoft/bedrock/issues/719


andrebriggs (changes requested):
Please update `validations.sh` in `tests\` to not use the mono-repo flag

mtarng (approved):

bnookala (approved):

andrebriggs (approved):

PR #143, opened 2019-11-15
Hld pipeline flags and rename

### LINKED ISSUE #783: "spk hld install-manifest-pipeline" should take command like args as overrides of spk config
**As a:**
Developer I want to be able to specify args that are necessary to run `spk hld install-manifest-pipeline`

According to the codebase the follow items from the spk config file are used
```
     const orgName = config.azure_devops.org!;
      const pat = config.azure_devops.access_token!;
      const hldRepo = config.azure_devops.hld_repository!;
      const project = config.azure_devops.project!;
      const manifestRepo = config.azure_devops.manifest_repository!;
      const hldRepoName = "HLD";
```

I would like to see arg flags for each one of these 

**Acceptance Criteria:**
I can run this command with a SPK config gile

Yes. Update [here](https://github.com/andrebriggs/spk/blob/master/docs/hld-management.md#install-manifest-pipeline)

mtarng:
Closes https://github.com/microsoft/bedrock/issues/783

Also updates the name of the hld repository pipeline yaml to be more descriptive.

Also updates config to use https as the repo urls.

Also adds integration tests for hld pipelines (these pipelines will fail, because there's nothing inside the hld repo)


docs/hld-management.md
```diff
@@ -42,24 +42,24 @@ An example configuration is as follows:
 
 ```
 azure_devops:
-  org: "https://dev.azure.com/bhnook"
-  project: "fabrikam"
-  hld_repository: "dev.azure.com/bhnook/fabrikam/_git/hld"
-  manifest_repository: "dev.azure.com/bhnook/fabrikam/_git/materialized"
-  access_token: "sdjkhasdxa3dw4jkhgf4zyv7t" # This is your Personal Access Token with permission to modify and access both hld and manifest repositories.
+  access_token: "hpe3a9oiswgcodtfdpzfiek3saxbrh5if1fp673xihgc5ap467a" # This is your Personal Access Token with permission to modify and access this private repo. Leave this empty if project is public
```
mtarng, line 45: this is a fake PAT

evanlouie (approved):

bnookala (approved):

tests/validations.sh
```diff
@@ -80,6 +114,19 @@ cd ..
...
 # *** TODO: Get ride of duplication
 
+# First we should check hld pipelines exist. If there is a pipeline with the same name we should delete it
+hld_pipeline_exists $AZDO_ORG_URL $AZDO_PROJECT $hld_dir $manifests_dir
+
+# Create the hld to manifest pipeline
+echo "hld_dir $hld_dir"
+echo "hld_repo_url $hld_repo_url"
+echo "manifest_repo_url $manifest_repo_url"
+spk hld install-manifest-pipeline -o $AZDO_ORG -d $AZDO_PROJECT -p $ACCESS_TOKEN_SECRET -r $hld_dir -u https://$hld_repo_url -m https://$manifest_repo_url
```
andrebriggs, line 124: @mtarng why the `https://` in front of `$manifest_repo_url`? Is it required?

tests/README.md
```diff
@@ -135,6 +135,9 @@ pipeline.
    export ACCESS_TOKEN_SECRET=<b>REPLACE_ME</b>
    export ACR_NAME=<b>REPLACE_ME</b>
    export AZURE_DEVOPS_EXT_PAT=<b>REPLACE_ME</b>
+   export SP_PASS=<b>REPLACE_ME</b>
```
andrebriggs, line 138: @mtarng Do one of the `spk` methods expect these as an env var? I was thinking we didn't needs these just to run the shells script

tests/validations.sh
mtarng: Yeah, unfortunately, the node library [git-url-parse](https://www.npmjs.com/package/git-url-parse) that we're using to parse doesn't recognize something like `dev.azure.com/bhnook/fabrikam/_git/hld` as a valid git url, so we prepend `https://` to the url. That's also why i've updated the default spk-config.yaml values for `azure_devops` repo values.

tests/README.md
mtarng: https://github.com/CatalystCode/spk/blob/master/tests/validations.sh#L144

Seems like they're needed to create the variable groups, unless i'm missing something here.

tests/validations.sh
andrebriggs: @mtarng This seems like a bug we should fix. This is more overhead for users. If a bug doesn't exist I can create one

tests/README.md
andrebriggs: Got it. Thanks!

PR #144, opened 2019-11-15
Lower casing container image names

### LINKED ISSUE #782: SPK doesn't handle mixed case docker image names in service and pipeline creation
**Describe the bug:**
SPK allows mixed cases service names and those service names are used in the scaffolded `azure-pipelines.yaml` file. This is a problem because Docker images need to be lowercase.
 
**To Reproduce:**
Use SPK:

```
$ spk project init -m -d $services_dir
```

```
$ spk service create $FrontEnd -d $services_dir 
```
Where $FrontEnd is Fabrikam.Acme.FrontEnd

Then run 
```
# spk service install-build-pipeline -o $AZDO_ORG -r $mono_repo_dir -u $remote_repo_url -d $AZDO_PROJECT -l $services_dir -p $ACCESS_TOKEN_SECRET -v $FrontEnd
```

The Build agent output is:

<pre>
========================== Starting Command Output ===========================
/bin/bash --noprofile --norc /home/vsts/work/_temp/9b199997-b6c1-445c-b956-75ddc78692b1.sh
az acr build -r andrebriggs --image FabrikamAcmeFrontEnd:master-20191114.2 .
WARNING: Packing source code into tar to upload...
WARNING: Excluding '.gitignore' based on default ignore rules
WARNING: Uploading archived source code from '/tmp/build_archive_1fa3fd09d17649be93b048ce10158759.tar.gz'...
WARNING: Sending context (1.364 KiB) to registry: andrebriggs...
ERROR: Some of the properties of 'runRequest' are invalid.. InnerErrors: DockerBuildRequest:Image name '<b>Fabrikam.Acme.FrontEnd</b>:master-20191114.2' is invalid. 
<b><i>The acceptable patterns are 'repository' or 'repository:tag'.The repository and tag names follow the standardized docker repository and tag naming conventions. Tag templates are supported within '{{' and '}}'.</i> </b>
Please refer to docs for more information on using image name and tag:  https://aka.ms/az/acr/az-acr-build
<b>##[error]Bash exited with code '1'.</b>
Finishing: ACR Build and Publish
</pre>

The build will succeed if I manually change <b>Fabrikam.Acme.FrontEnd</b> to <b>fabrikam.acme.frontend</b>

**Expected behavior:**
That the build succeeds. Perhaps we should lowercase the docker name in the scaffolded `azure-pipelines.yaml` file.

Probably

bnookala:
Fixes https://github.com/microsoft/bedrock/issues/782 by lower-casing the container image name that is to be built.

### DIFF
--- a/src/lib/fileutils.ts
+++ b/src/lib/fileutils.ts
@@ -130,13 +130,13 @@ export const starterAzurePipelines = async (opts: {
                     ? "-" + projectPathParts.slice(-1)[0]
                     : "";
 
-                const lowerCaseProjectName = projectName.toLowerCase();
-
                 return {
                   script: generateYamlScript([
+                    `export BUILD_REPO_NAME=$(echo $(Build.Repository.Name)-${projectName} | tr '[:upper:]' '[:lower:]')`,
+                    `echo "Image Name: $BUILD_REPO_NAME"`,
                     `cd ${projectPath}`,
-                    `echo "az acr build -r $(ACR_NAME) --image $(Build.Repository.Name)${lowerCaseProjectName}:$(Build.SourceBranchName)-$(Build.BuildNumber) ."`,
-                    `az acr build -r $(ACR_NAME) --image $(Build.Repository.Name)${lowerCaseProjectName}:$(Build.SourceBranchName)-$(Build.BuildNumber) .`
+                    `echo "az acr build -r $(ACR_NAME) --image $BUILD_REPO_NAME:$(Build.SourceBranchName)-$(Build.BuildNumber) ."`,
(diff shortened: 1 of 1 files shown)


andrebriggs (commented):
@bnookala looks good. I'm thinking about if the smoke test is the only test we can have for this. Do you think a unit test make sense here?

evanlouie (approved):

bnookala:
On further consideration - it looks like the repository name can be capitalized too, which presents another place where capitalization strikes again for docker image names

mtarng (approved):

PR #147, opened 2019-11-18
Adding component yaml if one does not exist to HLD

### LINKED ISSUE #790: spk hld init should create a component.yaml
**Describe the bug:**

If we have an empty HLD, the HLD to manifest pipeline will fail to run

**To Reproduce:**

`spk hld init`
`git push`
`spk hld install-manifest-pipeline`
`git commit --allow-empty`
`git push`

HLD to manifest pipeline will fail because hld init does not output a component.yaml

**Expected behavior:**

HLD to manifest pipeline should succeed

**Additional context:**

bnookala:
Fixes https://github.com/microsoft/bedrock/issues/790#issuecomment-554829305

### DIFF
--- a/src/commands/hld/init.test.ts
+++ b/src/commands/hld/init.test.ts
@@ -32,8 +32,8 @@ describe("Initializing an HLD Repository.", () => {
     expect(fs.existsSync(randomTmpDir)).toBe(true);
 
     // Verify new azure-pipelines created
-    const filepaths = ["manifest-generation.yaml"].map(filename =>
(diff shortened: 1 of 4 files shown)


edaena (approved):

mtarng (approved):
Discussed about having a type for Component.Yaml's with @bnookala . Since SPK will not actually be ever loading or modifying a component.yaml, we're not going to create a type for it; all the modifications for component.yaml should be handled by fabrikate.

PR #149, opened 2019-11-19
Extending smoke test validations for builds and pull requests 

andrebriggs:
Previous PR #146 was merged pre-maturely. Additional features here:

- We now check to make sure pipeline builds actually succeed.
- We merge automated  pull requests (that kick off more pipeline builds)
- Expanding `SP_TENANT` env var
- Publishing smoke test log as an artifact for easier debugging on AzDO


tests/validations.sh
```diff
@@ -190,8 +192,12 @@ pipeline_exists $AZDO_ORG_URL $AZDO_PROJECT $FrontEnd
 spk service install-build-pipeline -o $AZDO_ORG -r $mono_repo_dir -u $remote_repo_url -d $AZDO_PROJECT -l $services_dir -p $ACCESS_TOKEN_SECRET -v $FrontEnd  >> $TEST_WORKSPACE/log.txt
 
 # Verify the pipeline was created
-pipeline_created=$(az pipelines show --name $FrontEnd-pipeline --org $AZDO_ORG_URL --p $AZDO_PROJECT)
-# TODO: Verify the pipeline run was successful
+pipeline_name="$FrontEnd-pipeline"
```
mtarng: We can pass this name into line 192 with `-n` so that we can always be sure the name will match.

tests/validations.sh
```diff
@@ -129,7 +129,9 @@ spk hld install-manifest-pipeline -o $AZDO_ORG -d $AZDO_PROJECT -p $ACCESS_TOKEN
 
 # Verify the pipeline was created
 pipeline_created=$(az pipelines show --name $hld_dir-to-$manifests_dir --org $AZDO_ORG_URL --p $AZDO_PROJECT)
-# TODO: Verify the pipeline run was successful
+
+# Verify the pipeline run was successful
+verify_pipeline_with_poll $AZDO_ORG_URL $AZDO_PROJECT $hld_dir-to-$manifests_dir 180 15
```
yradsmikham, line 134: Is there a reason why the "poll interval" and "poll timeout" are specified as 180 and 5, respectively? How did you determine these numbers?

yradsmikham (approved):

tests/validations.sh
andrebriggs: @yradsmikham From trial and error around how long it usually takes

tests/validations.sh
andrebriggs: @mtarng addressed

mtarng (approved):

PR #159, opened 2019-11-27
Added support for remote backend in `spk infra generate`

### LINKED ISSUE #728: Provide support for passing backend remote state using `spk infra generate`
**As a:**
Developer

**I want:**
`spk infra generate` to recursively apply remote state configurations for tfstate files from parent directory to sub-directories for terraform deployment

**So that:**
at the final sub-directory, it creates a generated directory and fills the backend tfvars using the backend state variables accumulated from parent and child levels.

Upon generation, `spk infra generate` should:

- [x] Check in (parent) current working directory and verifies if it contains a definition.json:
- [x] If so, check the definition for any specification of a backend state. Then, iterate through child directory and create a backend.tfvars file

**Acceptance Criteria:**

When a user runs `spk infra generate east` for example, this should read in the backend configuration in the `definition.json` of the (parent) current working directory, and then apply the backend configuration of the `definition.json` of the `east` cluster. It creates a `generated` directory, with a `backend.tfvars` with the accumulated variables.

Yes.

### LINKED ISSUE #828: `definition.json` requires a "variable" block or `spk infra generate` will fail
**Describe the bug:**
In a multi-cluster set-up, if the parent or root `definition.json` file does not specify a "variable" block then `spk infra generate` will fail with `There was an error with generating the spk tfvars object.`

**To Reproduce:**
Use a parent `definition.json` without a `variable` block.

**Expected behavior:**
`spk infra generate` should be handle generation even though there are no "shared" variables in the parent `definition.json`.

No.

yradsmikham:
Close https://github.com/microsoft/bedrock/issues/728.
Close https://github.com/microsoft/bedrock/issues/828. (Bug)

### DIFF
--- a/src/commands/infra/generate.ts
+++ b/src/commands/infra/generate.ts
@@ -12,7 +12,6 @@ import {
 import { logger } from "../../logger";
 import * as infraCommon from "./infra_common";
 import { copyTfTemplate } from "./scaffold";
-import { definitionForAzureRepoPipeline } from "../../lib/pipelines/pipelines";
 
 const git = simpleGit();
 
--- a/src/commands/infra/generate.ts
+++ b/src/commands/infra/generate.ts
@@ -12,6 +12,7 @@ import {
 import { logger } from "../../logger";
 import * as infraCommon from "./infra_common";
 import { copyTfTemplate } from "./scaffold";
+import { definitionForAzureRepoPipeline } from "../../lib/pipelines/pipelines";
 
 const git = simpleGit();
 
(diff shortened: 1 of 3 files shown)


src/commands/infra/generate.ts
```diff
@@ -200,49 +200,133 @@ export const generateConfig = async (projectPath: string): Promise<void> => {
...
+              if (parentKey === leafKey) {
+                let parentVal = parentDefinitionJSON.variables[parentKey];
+                parentVal = leafDefinitionJSON.variables[leafKey];
+              } else {
+                // Append to parent variables block
+                const leafVal = leafDefinitionJSON.variables[leafKey];
+                parentDefinitionJSON.variables[leafKey] = leafVal;
+              }
+            }
+          }
```
edaena: nit: this is the same logic as the under `Create a backend.tfvars for remote backend configuration`, it can be a method and you can pass either `parentDefinitioinJSON.variables` or `leafDefinitionJSON.backend`.

NathanielRose (approved):
Lgtm - working on Linux/windows.
The only bug is `/r` is being added to the backend block variables in windows.

```
  "backend": {
    "storage_account_name": "storage-account-name\r",
    "access_key": "storage-account-access-key\r",
    "container_name": "storage-account-container\r",
    "key": "tfstate-key\r"
  },
```

src/commands/infra/generate.ts
yradsmikham: I had meant to do that, thanks for the reminder! :)

yradsmikham:
> Lgtm - working on Linux/windows.
> The only bug is `/r` is being added to the backend block variables in windows.
> 
> ```
>   "backend": {
>     "storage_account_name": "storage-account-name\r",
>     "access_key": "storage-account-access-key\r",
>     "container_name": "storage-account-container\r",
>     "key": "tfstate-key\r"
>   },
> ```

Resolved! :)

edaena (approved):

PR #166, opened 2019-12-18
Added task in pipeline to fail build on code coverage regression

### LINKED ISSUE #860: Prevent code coverage regression in SPK
Overall line code coverage for `spk` has decreased by 4% between November and December
* 65.9% (1240 of 1881 lines)  to 61.7% (1398 of 2264 lines) 

Overall branch coverage has decreased by 7% in the same timeline:
* 56.5% (307 of 543) to 49.1% (362 of 736). 

Overall we only added ~1000 more lines of code.

The image is from Dec 6th:
image

A solution is fail the `spk` build when the code coverage regresses. This is not a panacea. Folks can just write poor tests to achieve a code coverage metric but those changes should be scrutinized by reviewers

andrebriggs:
Added a [Build Quality Checks](https://marketplace.visualstudio.com/items?itemName=mspremier.BuildQualityChecks#adding-the-task-to-a-yaml-build-definition) task to the `azure-pipelines.yaml`.

The configuration will compare branch coverage numbers from the previous build to the current one. If lower the build will failure. In turn GitHub PR checks will failure. 

Hopefully to this drive paying attention to testing. 

Closes https://github.com/microsoft/bedrock/issues/860

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -70,8 +70,8 @@ steps:
       checkCoverage: true
       coverageFailOption: 'Previous Value'
       coverageType: branches
-      buildConfiguration: '$(BuildConfiguration)'
-      buildPlatform: '$(BuildPlatform)'
+      buildConfiguration: ''
+      buildPlatform: ''
       explicitFilter: true
       baseDefinitionId: 128
       baseBranchRef: refs/heads/master
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -60,7 +60,7 @@ steps:
       artifact: spk_linux_node_12
     condition: and(eq(variables['Agent.JobStatus'], 'Succeeded'), endsWith(variables['Agent.JobName'], 'node_12_x'))
 
-  - task: mspremier.BuildQualityChecks.QualityChecks-task.BuildQualityChecks@5
+  - task: mspremier.BuildQualityChecks.QualityChecks-task.BuildQualityChecks@6
     displayName: 'Check build quality'
     inputs:
       checkWarnings: false
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -63,6 +63,9 @@ steps:
   - task: mspremier.BuildQualityChecks.QualityChecks-task.BuildQualityChecks@5
     displayName: 'Check build quality'
     inputs:
+      checkWarnings: false
+      warningFailOption: fixed
+      warningFilters: ''
       showStatistics: true
       checkCoverage: true
       coverageFailOption: 'Previous Value'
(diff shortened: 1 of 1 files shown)


evanlouie (approved):
I'm okay with this, although I would like to hear input from other people on the team. Code coverage is an overall decent measurement of how hardened we are, but enforcing a rule like this can potentailly make us start to focus on the wrong things.

andrebriggs:
> I'm okay with this, although I would like to hear input from other people on the team. Code coverage is an overall decent measurement of how hardened we are, but enforcing a rule like this can potentailly make us start to focus on the wrong things.

Agreed. I mentioned to @bnookala offline that we don't want to over index of one measure. The point is to get folks to pay more attention to the code coverage as a proxy to good engineering practices. 

We still need to review and make sure that unit test are **_actually_** unit tests and not file system touching, network calling integration tests that may get high code coverage but have poor code design.

jmspring:
> > I'm okay with this, although I would like to hear input from other people on the team. Code coverage is an overall decent measurement of how hardened we are, but enforcing a rule like this can potentailly make us start to focus on the wrong things.
> 
> Agreed. I mentioned to @bnookala offline that we don't want to over index of one measure. The point is to get folks to pay more attention to the code coverage as a proxy to good engineering practices.
> 
> We still need to review and make sure that unit test are **_actually_** unit tests and not file system touching, network calling integration tests that may get high code coverage but have poor code design.

In general, I prefer warnings to blocking gates, but I understand the intent and goal.  Is the comparison an absolute value or relative?  Say I add come code and it triggers a rounding error based on fraction.  Should that be a fail?

If one were to add debugging code that was only triggered under certain conditions (needed for insight, etc) but didn’t make sense to test, that could lower the coverage.

Not every line of testing (command line handling?) necessitates coverage by a test.

Again I understand the intent and goal and agree with it, but are there scenarios (like those mentioned above) that don’t make sense to test and could result in a failure in a checkin?

andrebriggs:
@jmspring I'm not sure on the specifics of the rounding but it should be consistent. All this task does is look at the current and previous code coverage. Coverage is performed by an existing task. Code coverage does not take into account things like comments and non-executable code.

The problem with warnings is that people ignore them in the _rush_ to get code checked in. in case by case scenarios we can easily override behavior and merge a PR.

sarath-p (approved):
Overall, I agree with others that it is one of the way to enforce code coverage which is required. I prefer reviewers make an exception when approving PR's with low code coverage instead of making build fail and merge with an exception.

jmspring (approved):
For now lgtm, discussion should continue

PR #167, opened 2019-12-18
Infrastructure Integration Test for Scaffold

NathanielRose:
- Added smoke test pipeline support for validating `spk infra scaffold`
- Added function for validating file contents
- Closes - https://github.com/microsoft/bedrock/issues/785


smoke-test-pipeline.yml
```diff
@@ -3,7 +3,7 @@ trigger: none
 schedules:
 - cron: "0 3 * * *"
   branches:
-    include: [ master ] 
+    include: [ master, nate.infra.integration.scaffold ]
```
andrebriggs: @NathanielRose why run off of a separate branch?

smoke-test-pipeline.yml
```diff
@@ -73,6 +77,14 @@ stages:
         # Get exit status
         status=$?
         [ $status -eq 0 ] && echo "Test Done!" || (echo "Test had issues" && exit 1)
+
+
+        # Run the infra test
+        bash ./infra-validations.sh
```
andrebriggs, line 82: I'm thinking we can run the infra-validations in a separate stage or job. See east and west in the image below. Both tests need SPK on the file system. We would need to set the env vars in each stage/job. Thoughts?

image

tests/README.md
```diff
@@ -94,20 +100,22 @@ If you wish to run these tests locally, skip ahead to
    following key/values:
    - AZDO_PROJECT (e.g. `bedrock`)
    - AZDO_ORG (e.g. `epicstuff`)
-   - AZDO_PAT (e.g. Personal Access Token with _read/write/manage_ access to
+   - AZDO*PAT (e.g. Personal Access Token with \_read/write/manage* access to
```
andrebriggs: Not sure what these changes are adding.

tests/infra-validations.sh
```diff
@@ -0,0 +1,103 @@
...
+
+if [ ! -d "$TEST_WORKSPACE" ]; then
+  echo "Directory '$TEST_WORKSPACE' does not exist"
+  mkdir $TEST_WORKSPACE
+  echo "Created '$TEST_WORKSPACE'"
+fi
+
+cd $TEST_WORKSPACE
+
+# Setup simple TF Tepmlate ------------------
```
andrebriggs: Spelling

tests/infra-validations.sh
```diff
@@ -0,0 +1,103 @@
...
+# Single Cluster scaffold for template
+pwd
+echo "../$terraform_template_dir"
+echo "$tf_template_version"
+spk infra scaffold -n $infra_hld_dir --source "$source" --version "$tf_template_version" --template "template"
+# Validate the definition in the Infra-HLD repo ------------------
+file_we_expect=("definition.yaml")
+validate_directory "$TEST_WORKSPACE/$infra_hld_dir" "${file_we_expect[@]}"
+# Validate the contents of the definition.yaml
+validate_file "$TEST_WORKSPACE/$infra_hld_dir/definition.yaml" $validation_test_yaml
```
andrebriggs: Since this is a yaml file with particular details in it are we planning to verify further or will the `spa infra generate` tests implicitly verify that?

andrebriggs (changes requested):

andrebriggs:
Also not sure why we have a separate pipeline for this?
image

smoke-test-pipeline.yml
NathanielRose: Being used for testing, will be removed on final merge.

tests/infra-validations.sh
NathanielRose: Since the definition.yaml will have the PAT tokens for the source, we are doing a check for tf contents within the inputs block of the yaml. We can expand this later and do more validations with generate.

NathanielRose:
> Also not sure why we have a separate pipeline for this?
> image

Moving forward when modifying integration tests, the best way to reduce the creation of resources is to:
- fork then validate in your own project 
- then PR to a master where a pipeline is triggered based on integration test modifications and builds using the forked branch.

1 caveat is how to assure the variable groups in master will work with new PR

This pipe will be deleted, it was for testing

smoke-test-pipeline.yml
yradsmikham: This makes sense. As long as they're independent of each other, it would be efficient to have them run in parallel.

tests/infra-validations.sh
yradsmikham: @andrebriggs the `validate_file` function is sufficient in validating the contents of the yaml file. This will be re-used to validate contents in the `spk.tfvars` file in the `spk infra generate` integration test. For now, I think this is sufficient for `spk infra scaffold`.

smoke-test-pipeline.yml
```diff
@@ -58,12 +62,18 @@ stages:
...
         export SPK_LOCATION="$(pwd)/spk-linux"
         export ACCESS_TOKEN_SECRET=$(AZDO_PAT)
         export SP_PASS=$(SP_PASS)
         export SP_TENANT=$(SP_TENANT)
 
+        # Add to downstreams (Needs Fix)
+        echo "##vso[task.setvariable variable=SPK_LOCATION]$SPK_LOCATION"
+        echo "##vso[task.setvariable variable=ACCESS_TOKEN_SECRET]$ACCESS_TOKEN_SECRET"
+        echo "##vso[task.setvariable variable=SP_PASS]$SP_PASS"
+        echo "##vso[task.setvariable variable=SP_TENANT]$SP_TENANT"
```
yradsmikham: Is there a reason for echoing these variables?

yradsmikham (dismissed):

smoke-test-pipeline.yml
NathanielRose: Captured here - https://github.com/microsoft/bedrock/issues/863
Experiencing issues with current bedrock Integration tests in this method. May require refactoring.

tests/README.md
yradsmikham: If this is caused by husky, might be good to disable temporarily.

NathanielRose: Resolved

tests/infra-validations.sh
andrebriggs: When I looked yesterday the `validate_file` function didn't look like it did that much

yradsmikham: All the `validate_file` function is doing is scanning a file via `grep` for a particular string. This is how we validate whether a file (e.g. `definition.yaml` or `spk.tfvars`) was scaffolded or generated correctly.

andrebriggs: @yradsmikham I already I mentioned this to @NathanielRose 
> The check that you have is checking correctly. All that I mean is that we are not doing any sort of YAML file parsing. A grep search would succeed if the file was json internally. 

My assumption is that test for 'spk infra generate' will do that for us. We are on the same page

yradsmikham: Gotcha, yes, you're correct. If `spk infra generate` is capable of generating a `spk.tfvars` from the `definition.yaml` from scaffold, then this will be validated.

smoke-test-pipeline.yml
yradsmikham: Great, I can try to make this work when I add the integration tests for `spk infra generate`.

smoke-test-pipeline.yml
andrebriggs: @yradsmikham @NathanielRose please remove

tests/README.md
```diff
@@ -14,12 +14,18 @@ tests.
 
 - This file has simple functions that are reused by `validations.sh`.
 
+`infra-validations.sh`
+
+- This file orchestrates infrastructure scenarios and is design to fail on first
```
edaena: nit: "is designed" instead of "is design"

andrebriggs (approved):
Approved. As mentioned before I'm not a fan of the infra test being a sequential test after `spk services` validation. We will have to refactor in the future as @NathanielRose  mentioned.

tests/infra-validations.sh
```diff
@@ -0,0 +1,102 @@
...
+git add -A
+
+# See if the remote repo exists
+repo_exists $AZDO_ORG_URL $AZDO_PROJECT $terraform_template_dir
+
+# Create the remote repo for the local repo
+created_repo_result=$(az repos create --name "$terraform_template_dir" --org $AZDO_ORG_URL --p $AZDO_PROJECT) >> $TEST_WORKSPACE/log.txt
+
+# Extract out remote repo URL from the above result
+remote_repo_url=$(echo $created_repo_result | jq '.remoteUrl' | tr -d '"' )
```
edaena, line 74: what happens if the call above fails?
`az repos create`

yradsmikham: Then the integration test will fail. This is also how `spk services` is executing their integration tests; they rely on the `az cli` to create their mono, manifests and HLD repos.

edaena (approved):

evanlouie (approved):
lgmt

mtarng:
Looks good to me. Question though, should we run `validation.sh` and `infra-validations.sh` in different pipelines? It'd isolate them, and let them run in parallel. Just thoughts

mtarng (approved):

yradsmikham:
> Looks good to me. Question though, should we run `validation.sh` and `infra-validations.sh` in different pipelines? It'd isolate them, and let them run in parallel. Just thoughts

@mtarng Yes! We do plan on doing that. We created an issue for this: https://github.com/microsoft/bedrock/issues/863

PR #168, opened 2019-12-18
Add more tests for storage

edaena:


edaena:
Coverage up from 33.3% to 67.3%

src/lib/azure/storage.ts
```diff
@@ -149,7 +149,7 @@ export const getStorageAccountKeys = async (
     accountName
   );
 
-  if (typeof keysResponse.keys !== undefined) {
+  if (typeof keysResponse.keys !== "undefined") {
```
sarath-p, line 152: Did you try `!typeof keysResponse.keys` option?

sarath-p (approved):
Looks good. Added a one comment.

src/lib/azure/storage.ts
```diff
@@ -233,7 +233,8 @@ export const getStorageAccount = async (
       resourceGroup
     );
 
-    if (accounts === undefined || accounts === null) {
+    if (typeof accounts === "undefined" || accounts === null) {
+      logger.info("__no");
```
andrebriggs: @edaena What does `__no` mean?

andrebriggs (approved):

src/lib/azure/storage.ts
edaena: ah, I need to remove that useless log.

src/lib/azure/storage.ts
edaena: sync'ed offline. typeof returns a string, therefore the comparison needs to be to "undefined" not just undefined.

PR #169, opened 2019-12-19
Making hld argument fallback on config for project cvg

mtarng:
Minor change to fallback to config when creating a variable group.


bnookala (approved):

edaena (approved):

sarath-p (approved):
Looks good.

PR #170, opened 2019-12-23
Infrastructure Integration Test for Generate

### LINKED ISSUE #841: Integration Testing for `spk infra generate`
**As an:**
Operator

**I want:**
End-to-End testing that encapsulates `spk infra generate`

**So that:**
I can assure new features to infra that are out of scope for a unit test are thoroughly vetted before merged.
 
1. Download SPK
2. Write out simple definition.yaml file in a simple directory hierarchy
3. Execute `spk infra generate`
6. Verify creation generated Terraform files

**Acceptance Criteria:**
- [x] Add infra E2E integration test to AzDO pipeline
    - [x] Create a simple test repo for templates such as `azure-simple` or `azure-single-keyvault`
    - [x] Create another simple Infra HLD repo with basic definition files in a multi-cluster directory hierarchy
    - [x] Test generate function based on simple definition files to produce generated Terraform files

Yes (Integration Testing SPK doc)

### LINKED ISSUE #863: [SPK] Embed Integration Tests for Infra and Service validation into multi stage parallel pipeline
**As a:**
Operator

**I want:**
End-to-End testing that runs spk validations in parallel

**So that:**
Integration validation test are isolated on failure

"we can run the infra-validations in a separate stage or job. See east and west in the image below. Both tests need SPK on the file system. We would need to set the env vars in each stage/job."
image

**Additional context:**
Env Var issue also being experienced with Bedrock Pipeline for integration tests.

Yes

yradsmikham:
Close https://github.com/microsoft/bedrock/issues/841.
Close https://github.com/microsoft/bedrock/issues/863.

Infra Integration Tests are executed in a parallel job within pipeline.


edaena (approved):

smoke-test-pipeline.yml
```diff
@@ -78,16 +74,78 @@ stages:
...
+      env:
+        FUNC_SCRIPT: $(FUNC_SCRIPT)
+        TEST_SCRIPT2: $(TEST_SCRIPT2)
+    - bash: |
+        # Install the extension
+        export AZURE_DEVOPS_EXT_PAT=$(AZDO_PAT)
+        az extension add --name azure-devops
+
+        # Login to Azure
+        echo "az login --service-principal --username $(SP_APP_ID) --password $(SP_PASS) --tenant $(SP_TENANT)"
```
NathanielRose, line 122: @yradsmikham Do we need to login twice when we have multiple jobs?  `SPK_LOCATION` `ACCESS_TOKEN_SECRET` and other vars are being set twice. Could we move this to pipeline variables?

tests/README.md
```diff
@@ -86,7 +86,7 @@ tests.
 | -------------------------- | -------- |
 | spk infra scaffold         | ✅       |
 | spk infra validate onboard | 🚫       |
```
NathanielRose, line 88: @andrebriggs What is expected for the `spk infra validate onboard`?

tests/infra-validations.sh
```diff
@@ -100,3 +100,16 @@ file_we_expect=("definition.yaml")
...
+spk infra generate
+
+# Verify that the Terraform files generation was successful
+# Confirm that generated directory created, spk.tfvars created, and tf templates copied
+generated_directory="$TEST_WORKSPACE/$infra_hld_dir-generated"
+file_we_expect=("spk.tfvars" "main.tf" "variables.tf")
+validate_directory "$generated_directory" "${file_we_expect[@]}"
+
+# Confirm contents of the spk.tfvars file are correct
+validate_file "$generated_directory/spk.tfvars" 'rg_name = "<insert value>"'
```
NathanielRose, line 117: @yradsmikham One caveat I see is that we're not testing for backend or partial config on the generate at 2 levels instead of 1.

NathanielRose (approved):
LGTM - Minor Comments below

evanlouie (approved):

andrebriggs:
@yradsmikham do you have an example of the pipeline running? The reason I ask is because this doesn't look like to runs in parallel but rather sequential. Moreover it seems we are duplicating steps like `DownloadPipelineArtifact` and some of the bash steps. 

This means that even if it is running in parallel we risk overwriting the binary and scripts. This could have side effects. 

What was proposed [here](https://github.com/CatalystCode/spk/pull/167#discussion_r359563505) indicates that there are shared components to reduce duplication

smoke-test-pipeline.yml
yradsmikham: That's correct. When you have run them in separate jobs in parallel, it essentially spins up another agent. So you will need to download and install SPK, and reset other configurations for the agent in order to run the infra tests.

yradsmikham:
> @yradsmikham do you have an example of the pipeline running? The reason I ask is because this doesn't look like to runs in parallel but rather sequential. Moreover it seems we are duplicating steps like `DownloadPipelineArtifact` and some of the bash steps.
> 
> This means that even if it is running in parallel we risk overwriting the binary and scripts. This could have side effects.
> 
> What was proposed [here](https://github.com/CatalystCode/spk/pull/167#discussion_r359563505) indicates that there are shared components to reduce duplication

@andrebriggs, please visit [here](https://dev.azure.com/yvradsmi/spk-infra/_build/results?buildId=963&view=results). I have confirmed that the jobs run in parallel. 

Screen Shot 2019-12-28 at 10 16 11 AM

https://github.com/CatalystCode/spk/pull/170#discussion_r361803067

It won't be overwriting scripts because they are independent of each other. Running jobs in parallel here means they are running in separate agents.

tests/infra-validations.sh
yradsmikham: Not a problem, we can add backend configuration to the yaml file and test for it.

smoke-test-pipeline.yml
yradsmikham: To reduce some of the preparation, it might make sense to create a custom build agent for SPK, so that we do not have to repeat some of these steps.

andrebriggs (approved):

PR #177, opened 2020-01-09
Middleware Tracking/Injection for Services

### LINKED ISSUE #779: As a spk user, I would like the ability to utilize middlewares in my ingress routes
**As a:**

spk user

**I want:**

A way to specify middlewares in my bedrock.yaml and an exposed method on the `spk` tool that allows me to specify middlewares

**So that:**

A generated ingress-route created via `spk hld reconcile` will also use the middlewars specified in the bedrock.yaml

`spk service create` to take a list of middlewares and `spk hld reconcile` to output ingress routes that use the middlewares

**Acceptance Criteria:**

- [ ] `spk service create` should accept a list of (optional) middlewares names (as strings) that should be written to the bedrock.yaml file ie:

```
rings:
  master:
    isDefault: true
services:
  ./root-service:
    middlewares:
     - foo
     - bar
     - baz
    helm:
      chart:
        branch: 'master'
        git: 'https://github.com/helm/charts.git'
        path: 'stable/airflow'
variableGroups:
  - nov11
```

- [ ] `spk hld reconcile` should take the middlewares written by `spk service create` to the bedrock.yaml file, and output an ingress route with middlewares:

```
kind: IngressRoute
metadata:
  name: root-service-master-route
spec:
  entryPoints:
    - web
  routes:
  - match: PathPrefix(`/root-service`) && Headers(`Ring`, `master`)
    kind: Rule
    services:
    - name: root-service-master
       port: 8000
    middlewares:
    - name: foo
```

None

**Additional context:**

We need to be able to provide middlewares for an ingress route, and `bedrock.yaml` is the best extensibility point for that

See: https://docs.traefik.io/routing/providers/kubernetes-crd/#middleware for more documentation on how middlewares work with the k8s crds.

Yes

evanlouie:
Added support for tracking and injecting Traefik2 middlewares on a per-service
basis.

- Services tracked in `bedrock.yaml` now supports an optional `middlewares` key
  that accepts a list of Traefik2 middlewares.
  - These middlewares are injected into the outputted Traefik2 `IngressRoutes`
    generated from `spk hld reconcile`.
- Fixed broken user input validations for `variableGroupName`.
- Fixed `build` script in `package.json`. Was breaking on hosts that use Node 13
  as pkg currently does not have prebuilt binaries for versions `>12`

closes https://github.com/microsoft/bedrock/issues/779

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "contributors": [],
   "license": "MIT",
   "scripts": {
-    "build": "shx rm -rf dist && webpack && pkg dist/spk.js && shx mv spk-{linux,macos,win.exe} dist",
+    "build": "shx rm -rf dist && webpack && pkg -t node12-linux-x64,node12-macos-x64,node12-win-x64 dist/spk.js && shx mv spk-{linux,macos,win.exe} dist",
     "lint": "tslint 'src/**/*.ts{,x}'",
     "lint-fix": "tslint --fix 'src/**/*.ts{,x}'",
     "test": "jest --reporters=jest-junit --reporters=default --coverage --coverageReporters=cobertura --coverageReporters=html",
--- a/docs/service-management.md
+++ b/docs/service-management.md
@@ -38,17 +38,19 @@ Usage: service create|c [options] <service-name>
 Add a new service into this initialized spk project repository
 
 Options:
-  -c, --helm-chart-chart <helm-chart>            bedrock helm chart name. --helm-chart-* and --helm-config-* are exclusive; you may only use one. (default: "")
-  -r, --helm-chart-repository <helm-repository>  bedrock helm chart repository. --helm-chart-* and --helm-config-* are exclusive; you may only use one. (default: "")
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -6,7 +6,7 @@
   "contributors": [],
   "license": "MIT",
   "scripts": {
-    "build": "shx rm -rf dist && webpack && pkg dist/spk.js && shx mv spk-{linux,macos,win.exe} dist",
+    "build": "shx rm -rf dist && webpack && pkg -t node12-linux-x64,node12-macos-x64,node12-win-x64 dist/spk.js && shx mv spk-{linux,macos,win.exe} dist",
```
evanlouie, line 9: @andrebriggs note this. The `build` script now always try's to build on a node 12 binary instead of whatever is on the host machine. To my knowledge, this shouldn't effect any of the the integration tests.

andrebriggs: @evanlouie That should be fine. the Node 12 build produces the build artifact.

sarath-p (approved):
Looks good.

yradsmikham (approved):

PR #179, opened 2020-01-10
Resolving `spk infra generate` bugs

### LINKED ISSUE #856: Leaf definition.json requires template and source
As an operator of a large number of "largely the same" clusters I would like to be able to centralize as much of the common parameters as possible.

When defining a cluster, it is not currently possible to have a `template`, `source`, and `version` centrally defined in a higher in the directory hierarchy `definition.json` and omit these in the leaf `definition.json`.  Doing so results in an error that looks like:

```
$ spk infra generate --project prod-blue
info:    spk will generate the following project: prod-blue
info:    Project folder found. Extracting information from definition.json files.
error:   Unable to validate project folder definition.json file. Is it malformed?
error:   Error occurred while generating project deployment files
error:   Cannot read property 'replace' of undefined
```

### LINKED ISSUE #870: spk infra generate fails when leaf definition.yaml is not present
**Describe the bug:**
When there exists a parent `definition.yaml` and a leaf directory without a leaf `definition.yaml`, spk infra generate fails to fall back on parent definitions to generate Terraform files.

Example setup:
```
|- discovery-service
    |- definition.yaml
    |- east/
```
**To Reproduce:**

Create a hierarchy of directories like the one shown above, and execute `spk infra generate -p east` in the `discovery-service` directory.
**Expected behavior:**

**Additional context:**
`spk infra generate` should be able to recognize that a parent `definition.yaml` exists and fall back on the parent definitions where a leaf `definition.yaml` does not exist.

No.

yradsmikham:
Closes https://github.com/microsoft/bedrock/issues/856.
Closes https://github.com/microsoft/bedrock/issues/870.

### DIFF
--- a/src/commands/infra/generate.ts
+++ b/src/commands/infra/generate.ts
@@ -193,7 +193,7 @@ export const validateRemoteSource = async (
       logger.warn(
         `Provided source in template directory was not found, attempting to clone the template source repo locally.`
       );
-      mkdirp.sync(sourcePath);
+      await createGenerated(sourcePath);
     } else {
       logger.info(
         `Source template folder found. Validating existence of repository.`
--- a/src/commands/infra/generate.ts
+++ b/src/commands/infra/generate.ts
@@ -62,7 +62,8 @@ export const generateCommandDecorator = (command: commander.Command): void => {
 /**
  * Checks if definition.yaml is present locally to provided project path
  *
- * @param projectPath Path to the definition.yaml file
+ *  @param parentPath Path to the parent definition.yaml file
+ *  @param projectPath Path to the leaf definition.yaml file
  */
 export const validateDefinition = async (
   parentPath: string,
@@ -176,7 +177,7 @@ export const validateTemplateSources = async (
 /**
  * Checks if provided source, template and version are valid. TODO/ Private Repo, PAT, ssh-key agent
  *
- * @param projectPath Path to the definition.yaml file
+ * @param definitionJSON definition object in JSON format
  */
 export const validateRemoteSource = async (
   definitionJSON: string[]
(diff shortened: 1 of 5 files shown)


src/commands/infra/generate.ts
```diff
@@ -60,61 +65,111 @@ export const generateCommandDecorator = (command: commander.Command): void => {
  * @param projectPath Path to the definition.yaml file
  */
```
edaena, line 67: missing `parentPath` in the function doc

src/commands/infra/generate.ts
```diff
@@ -60,61 +65,111 @@ export const generateCommandDecorator = (command: commander.Command): void => {
...
+    // Check for parent definition.yaml and child definition.yaml
+    if (
+      fs.existsSync(path.join(parentPath, "definition.yaml")) &&
+      fs.existsSync(path.join(projectPath, "definition.yaml"))
+    ) {
+      logger.warn(
+        `A definition.yaml was found in: ${parentPath}, and was found in ${projectPath}`
       );
-      return false;
+      return "A";
```
edaena, line 83: nit: this can return something more descriptive like ParentLeafConfig instead of A and ParentConfig instead of B, just some name suggestions. It can be an enum.

samiyaakhtar (approved):

andrebriggs (changes requested):
@yradsmikham why do we have a `definition.yaml` and `definition.json`?

yradsmikham:
> @yradsmikham why do we have a `definition.yaml` and `definition.json`?

Deleted.

andrebriggs (approved):

src/commands/infra/generate.ts
yradsmikham: I think I am going to leave the configuration convention the way it is for now. Currently, the code docs explain what each configuration means, and this is actually only useful to the developer since it is not explicitly printed anywhere in the logs. When configurations become more complex, I will def keep this mind. Thanks!

PR #185, opened 2020-01-15
Adding coverageDeltaType to percentage

andrebriggs:
Previously we weren't setting the `coverageDeltaType`. The `percentage` value will compare between the current and previous code coverage value.


mtarng (approved):

evanlouie (approved):

jmspring (approved):
Lgtm

PR #186, opened 2020-01-15
Allow skipping ingress route creation

### LINKED ISSUE #831: Allow disabling ingress route creation
**As a:**

SPK user

**I want:**

To disable ingress route creation for services via bedrock.yaml

**So that:**

Select services do not have ingress routes generated. This could be useful in the event that I have a service that I want to consume internally (within the k8s cluster).

bedrock.yaml to be augmented to look like: 

```
rings:
  dev:
    isDefault: true
  prod:
services:
  homepage:
    port: 8001
  database:
    disableIngressRoute: true
```

**Acceptance Criteria:**

- [ ] If I set `disableIngressRoute` to true in my bedrock.yaml, I should expect that no ingress route mapping is generated as a consequence

None

**Additional context:**

We should allow the option of "opting out" of ingress route creation

Yes

bnookala:
This got to be a bit bigger than the requested task, but it does add much-needed tests for reconcile:

- Allows for skipping ingress route creation, when specified in bedrock.yaml
- Adds tests for reconcile
- Refactor much of the reconcile functionality to be more modular

Fixes https://github.com/microsoft/bedrock/issues/831

### DIFF
--- a/src/commands/hld/reconcile.ts
+++ b/src/commands/hld/reconcile.ts
@@ -121,7 +121,7 @@ export const reconcileHldDecorator = (command: commander.Command): void => {
           absHldPath
         );
       } catch (err) {
-        logger.error(`Error occurred while reconciling HLD`);
+        logger.error(`An error occurred while reconciling HLD`);
         logger.error(err);
         process.exit(1);
       }
@@ -247,7 +247,7 @@ export const execAndLog = async (commandToRun: string) => {
   if (commandResult.stderr) {
     logger.error(commandResult.stderr);
     throw Error(
-      `Error occurred when invoking command: ${commandResult.stderr}`
+      `An error occurred when invoking command: ${commandResult.stderr}`
     );
   }
 };
@@ -398,7 +398,7 @@ export const testAndGetAbsPath = (
 ): string => {
   const absPath = path.resolve(possiblyRelativePath);
   if (!test("-e", absPath) && !test("-d", absPath)) {
-    throw new Error(`Error: could not validate ${pathType} path.`);
+    throw new Error(`Could not validate ${pathType} path.`);
   }
   log(`Found ${pathType} at ${absPath}`);
   return absPath;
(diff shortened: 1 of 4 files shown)


evanlouie (approved):

src/commands/hld/reconcile.ts
```diff
@@ -8,12 +8,67 @@ import shelljs from "shelljs";
...
+  TraefikMiddleware
+} from "../../lib/traefik/middleware";
 import { logger } from "../../logger";
 
 const exec = promisify(child_process.exec);
 
-import { IBedrockFile } from "../../types";
+import { IBedrockFile, IBedrockServiceConfig } from "../../types";
+
+//
```
mtarng: ? this just a trailing comment?

bnookala: seems misc, will remove

mtarng (approved):

sarath-p (approved):
Looks good

src/commands/hld/reconcile.ts
```diff
@@ -223,3 +251,164 @@ const execAndLog = async (commandToRun: string) => {
...
+  }
+  log(`Found ${pathType} at ${absPath}`);
+  return absPath;
+};
+
+export const checkForFabrikate = (which: (path: string) => string) => {
+  const fabrikateInstalled = which("fab");
+  if (fabrikateInstalled === "") {
+    throw new Error(
+      `Error: Fabrikate not installed. Please fetch and install the latest version: https://github.com/microsoft/fabrikate/releases`
```
evanlouie: don't think we should prefix the errors with "Error: ". When an `Error` gets serialized via `.toString()` it automatically gets an "Error:" prefix. So this would result in `Error: Error: Fabrikate not installed...`

evanlouie (changes requested):

src/commands/hld/reconcile.ts
bnookala: makes sense - will change

src/commands/hld/reconcile.ts
```diff
@@ -223,3 +251,164 @@ const execAndLog = async (commandToRun: string) => {
...
+
+export const testAndGetAbsPath = (
+  test: (flags: TestOptions, path: string) => boolean,
+  log: (logline: string) => void,
+  possiblyRelativePath: string,
+  pathType: string
+): string => {
+  const absPath = path.resolve(possiblyRelativePath);
+  if (!test("-e", absPath) && !test("-d", absPath)) {
+    throw new Error(`Error: could not validate ${pathType} path.`);
```
evanlouie: no "Error:" prefix in message 🙏

evanlouie (approved):

mtarng (approved):

PR #189, opened 2020-01-17
[REFACTORING] replace any with interface

dennisseah:
mainly to avoid using `any` in the code base

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -132,15 +132,15 @@ export interface IServiceEndpointData {
   tenant_id: string;
 }
 
+export interface IVariableGroupDataVariable {
+  [key: string]: AzureKeyVaultVariableValue;
+}
+
 export interface IVariableGroupData {
   name: string;
   description: string;
   type: string;
-  variables: [
-    {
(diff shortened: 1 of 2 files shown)


bnookala (approved):

src/types.d.ts
```diff
@@ -132,15 +132,15 @@ export interface IServiceEndpointData {
...
 export interface IVariableGroupData {
   name: string;
   description: string;
   type: string;
-  variables: [
-    {
-      [key: string]: AzureKeyVaultVariableValue;
-    }
-  ];
+  variables: [IVariableGroupDataVariable];
```
evanlouie: I think this is supposed to be `variables: IVariableGroupDataVariable[]`

evanlouie: @sarath-p ^^ is that correct?

dennisseah: @evanlouie, you are right. I will change and test again. thanks

evanlouie (approved):

sarath-p (approved):
Looks good

PR #192, opened 2020-01-18
[REFACTORING] making code more compact

dennisseah:
1. move command description to a json file
2. create a command builder class
3. make `validateRequiredArguments` function non `async`
4. remove unused code in test file
5. create helper function for checking missing values
6. remove `any` type with proper type
7. add types to function
8. use process.exit(0) and process.exit(1) for terminating command 

### DIFF
--- a/src/commands/infra/scaffold.ts
+++ b/src/commands/infra/scaffold.ts
@@ -9,6 +9,8 @@ import { IConfigYaml } from "../../types";
 import { validateRemoteSource } from "./generate";
 import * as infraCommon from "./infra_common";
 
+const DEFINITION_YAML = "definition.yaml";
+
 interface ICommandOptions {
   [key: string]: string;
 }
@@ -75,7 +77,7 @@ export const execute = async (
         path.join(sourcePath, opts.template, "variables.tf")
       );
       await scaffold(opts.name, opts.source, opts.version, opts.template);
-      removeTemplateFiles(opts.name);
+      await removeTemplateFiles(opts.name);
       exitFn(0);
     } catch (err) {
       logger.error("Error occurred while generating scaffold");
@@ -130,7 +132,7 @@ export const validateVariablesTf = async (
       return false;
     }
     logger.info(
-      `Terraform variables.tf file found. Attempting to generate definition.yaml file.`
+      `Terraform variables.tf file found. Attempting to generate ${DEFINITION_YAML} file.`
     );
   } catch (_) {
     logger.error(`Unable to validate Terraform variables.tf.`);
@@ -226,18 +228,18 @@ export const copyTfTemplate = async (
  *
  * @param envPath path so the directory of Terraform templates
  */
-export const removeTemplateFiles = (envPath: string) => {
+export const removeTemplateFiles = async (envPath: string) => {
   // Remove template files after parsing
-  fs.readdir(envPath, (err, files) => {
-    if (err) {
-      throw err;
-    }
-    for (const file of files) {
-      if (file !== "definition.yaml") {
-        fs.unlinkSync(path.join(envPath, file));
-      }
-    }
-  });
+  try {
+    const files = await fs.readdirSync(envPath);
+    files
+      .filter(f => f !== DEFINITION_YAML)
+      .forEach(f => {
+        fs.unlinkSync(path.join(envPath, f));
+      });
+  } catch (e) {
+    logger.error(`cannot read ${envPath}`);
+  }
 };
 
 /**
@@ -390,15 +392,12 @@ export const scaffold = async (
         );
         const definitionYaml = yaml.safeDump(baseDef);
         if (baseDef) {
-          fs.mkdir(name, (e: any) => {
-            const confPath: string = path.format({
-              base: "definition.yaml",
-              dir: name,
-              root: "/ignored"
-            });
-            fs.writeFileSync(confPath, definitionYaml, "utf8");
-            return true;
+          const confPath: string = path.format({
+            base: DEFINITION_YAML,
+            dir: name,
+            root: "/ignored"
           });
+          fs.writeFileSync(confPath, definitionYaml, "utf8");
         } else {
           logger.error(`Unable to generate cluster definition.`);
         }
--- a/src/commands/infra/scaffold.ts
+++ b/src/commands/infra/scaffold.ts
@@ -5,8 +5,85 @@ import yaml from "js-yaml";
 import path from "path";
 import { Config } from "../../config";
 import { logger } from "../../logger";
+import { IConfigYaml } from "../../types";
 import { validateRemoteSource } from "./generate";
 import * as infraCommon from "./infra_common";
+
+interface ICommandOptions {
+  [key: string]: string;
+}
+
+export const validateValues = (
+  config: IConfigYaml,
+  opts: ICommandOptions
+): boolean => {
+  if (
+    !config.azure_devops ||
+    !config.azure_devops.access_token ||
+    !config.azure_devops.infra_repository
+  ) {
+    logger.warn(
+      "The infrastructure repository containing the remote terraform template repo and access token was not specified. Checking passed arguments."
+    );
+
+    if (!opts.name || !opts.source || !opts.version || !opts.template) {
+      logger.error(
+        "You must specify each of the variables 'name', 'source', 'version', 'template' in order to scaffold out a deployment."
+      );
+      return false;
+    }
+    logger.info(
+      "All required options are configured via command line for scaffolding, expecting public remote repository for terraform templates or PAT embedded in source URL."
+    );
+  }
+  return true;
+};
+
+// Construct the source based on the the passed configurations of spk-config.yaml
+export const constructSource = (config: IConfigYaml) => {
+  const devops = config.azure_devops!;
+  const source = `https://spk:${devops.access_token}@${devops.infra_repository}`;
+  logger.info(
+    `Infrastructure repository detected from initialized spk-config.yaml.`
+  );
+  return source;
+};
+
+export const execute = async (
+  config: IConfigYaml,
+  opts: ICommandOptions,
+  exitFn: (status: number) => void
+) => {
+  if (!validateValues(config, opts)) {
+    exitFn(1);
+  } else {
+    opts.source = opts.source || constructSource(config);
+
+    try {
+      /* scaffoldDefinition will take in a definition object with a
+          null configuration. Hence, the first index is "" */
+      const scaffoldDefinition = ["", opts.source, opts.template, opts.version];
+      const sourceFolder = await infraCommon.repoCloneRegex(opts.source);
+      const sourcePath = path.join(infraCommon.spkTemplatesPath, sourceFolder);
+      await validateRemoteSource(scaffoldDefinition);
+      await copyTfTemplate(
+        path.join(sourcePath, opts.template),
+        opts.name,
+        false
+      );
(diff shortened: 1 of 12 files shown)


andrebriggs:
Thanks @dennisseah for the pull request. Currently the checks for your PR area failing because there is a drop in overall branch line coverage that is more than 1%.

> The code coverage value (58.4939%, 668 branches) is lower than the value of the previous build including variance! The previous build had a coverage value of 59.7964% (705 branches).

## Your PR at https://dev.azure.com/epicstuff/bedrock/_build/results?buildId=9807&view=codecoverage-tab
image

## Last commit to master at https://dev.azure.com/epicstuff/bedrock/_build/results?buildId=9797&view=codecoverage-tab
image

The links are a good starting point at looking at the coverage differences.

mtarng:
Seems like this could also be combined with this ticket https://github.com/microsoft/bedrock/issues/741
We are looking to standardize flag names and shorthand flags across all commands, though we don't have a design proposed yet. I like what you've done with the decorator.json file, but we'll probably want a central library for all commands to share/overlap. Command builder is great too.

Thoughts? @sarath-p @evanlouie @bnookala

dennisseah:
> Seems like this could also be combined with this ticket [microsoft/bedrock#741](https://github.com/microsoft/bedrock/issues/741)
> We are looking to standardize flag names and shorthand flags across all commands, though we don't have a design proposed yet. I like what you've done with the decorator.json file, but we'll probably want a central library for all commands to share/overlap. Command builder is great too.
> 
> Thoughts? @sarath-p @evanlouie @bnookala

I was combing thru the command and found that sometimes, we use --organization-name and other times, we use --org-name. Having in json, allow us to write code to scan for consistency. BTW, I found a bug, we have --build-script (like here, https://github.com/CatalystCode/spk/blob/getting-started-docs/src/commands/project/pipeline.ts#L45) and then we get the value via buildScriptUrl instead of buildScript (here https://github.com/CatalystCode/spk/blob/getting-started-docs/src/commands/project/pipeline.ts#L60)

mtarng:
yup. If you look at the ticket, there's a spreadsheet of all the current commands (https://microsoft.sharepoint.com/:x:/t/DevCrews-Episerver/EUA0JeUOgppHq_YY6sAkD68BYa3aHX1idc8cwOLgRT0gYg?e=sUo0jx)

Definitely want to get parity with all commands

andrebriggs (approved):

evanlouie:
was their a particular reason we chose to use json files for the decorators specs as opposed to exporting a JS object? I think some structural integrity would be added if we did that and had a shared interface for the decorators to export.

something like

```ts
interface ICommandDecorator {
  command: string;
  alias?: string;
  description: string;
  options?: {
    [flag: string]: {
      arg?: string;
      description: string;
      required: boolean;
    };
  };
}
```

src/commands/project/create-variable-group.ts
```diff
@@ -3,105 +3,112 @@ import commander from "commander";
...
+  } else {
+    try {
+      const { azure_devops } = Config();
+
+      const {
+        registryName,
+        servicePrincipalId,
+        servicePrincipalPassword,
+        tenant,
+        hldRepoUrl = azure_devops && azure_devops.hld_repository,
```
evanlouie: we should avoid using `&&` for de-structuring as we are now on TS 3.7.
Should use `hldRepoUrl = azure_devops?.hld_repository` instead

src/commands/project/create-variable-group.ts
```diff
@@ -3,105 +3,112 @@ import commander from "commander";
...
+        orgName = azure_devops && azure_devops.org,
```
evanlouie: Use `?.` operator instead of `&&`

src/commands/project/create-variable-group.ts
```diff
@@ -3,105 +3,112 @@ import commander from "commander";
...
+        personalAccessToken = azure_devops && azure_devops.access_token,
```
evanlouie: Use `?.` operator instead of &&

src/commands/project/create-variable-group.ts
```diff
@@ -3,105 +3,112 @@ import commander from "commander";
...
+        project = azure_devops && azure_devops.project
```
evanlouie: Use `?.` operator instead of &&

dennisseah:
> was their a particular reason we chose to use json files for the decorators specs as opposed to exporting a JS object? I think some structural integrity would be added if we did that and had a shared interface for the decorators to export.
> 
> something like
> 
> ```ts
> interface ICommandDecorator {
>   command: string;
>   alias?: string;
>   description: string;
>   options?: {
>     [flag: string]: {
>       arg?: string;
>       description: string;
>       required: boolean;
>     };
>   };
> }
> ```

Thanks for your comment, @evan

1. JSON is lightweight. We can parse it and generate documentation for the command line if we wish to (later in time)
2. during JSON import and calling the build function in CommandBuilder, the type checking is done. that’s i got an error if my json is incorrect (does not follow the interface)

src/lib/commandBuilder.test.ts
```diff
@@ -0,0 +1,100 @@
...
+        },
+        optionC: {
+          arg: "-c, --optionC <optionC>",
+          description: "description for optionC",
+          required: false
+        }
+      }
+    };
+
+    const options: ICommandOption[] = Object.getOwnPropertyNames(
```
evanlouie: should use `.keys` instead of `.getOwnPropertyNames`

src/lib/commandBuilder.test.ts
```diff
@@ -0,0 +1,100 @@
...
+        },
+        optionC: {
+          arg: "-c --optionC <optionC>",
+          description: "description for optionC",
+          required: true
+        }
+      }
+    };
+
+    const options: ICommandOption[] = Object.getOwnPropertyNames(
```
evanlouie: `.keys` instead of `.getOwnPropertyNames`

src/lib/commandBuilder.ts
```diff
@@ -0,0 +1,82 @@
...
+  command: commander.Command,
+  decorator: ICommandBuildElements
+): commander.Command => {
+  const cmd = command
+    .command(decorator.command)
+    .alias(decorator.alias)
+    .description(decorator.description);
+
+  const options = decorator.options;
+  Object.getOwnPropertyNames(options).forEach(name => {
```
evanlouie: `.keys` instead of `.getOwnPropertyNames`

src/lib/commandBuilder.ts
```diff
@@ -0,0 +1,82 @@
...
+ * @param decorator Descriptor for command building
+ * @param values Values to be inspected.
+ * @return error messages.
+ */
+export const validateForRequiredValues = (
+  decorator: ICommandBuildElements,
+  values: { [key: string]: string | undefined }
+): string[] => {
+  // gather the required variable names from the decorator options object
+  const requireds = Object.getOwnPropertyNames(decorator.options).filter(
```
evanlouie: `.keys` instead of `.getOwnPropertyNames`

src/lib/commandBuilder.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Command Option
+ */
+export interface ICommandOption {
+  arg: string;
+  description: string;
+  required: boolean;
+}
+
+/**
+ * Map of key (stirng) to its command option
```
evanlouie: typo (strirng => string)

evanlouie (changes requested):
some initial quick finds

src/lib/validator.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Values to be validated
+ */
+export interface IValidationValue {
+  value: undefined | null | string;
+  error: string;
+}
+
+/**
+ * Returns true of val is undefined, null or empty string.
```
evanlouie: I believe this doc is incorrect. Should be `Returns true if val is not undefined, not null, and not an empty string.`

src/lib/validator.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @param val Value to inspect
+ */
+export const hasValue = (val: undefined | null | string): boolean => {
```
evanlouie: function should be changed to a type-guard

```ts
export const hasValue = (val: undefined | null | string): val is string => {
  return val !== undefined && val !== null && val !== "";
};
```

evanlouie (changes requested):

src/commands/project/create-variable-group.ts
dennisseah: ok. thinking not to change the initial code too much

src/commands/project/create-variable-group.ts
dennisseah: ok. thinking not to change the initial code too much

src/commands/project/create-variable-group.ts
dennisseah: ok. thinking not to change the initial code too much

src/commands/project/create-variable-group.ts
dennisseah: ok. thinking not to change the initial code too much

src/lib/commandBuilder.test.ts
dennisseah: refactor the code and this is not needed now. thanks

src/lib/commandBuilder.test.ts
dennisseah: refactor the code and this is not needed now. thanks

src/lib/commandBuilder.ts
dennisseah: refactor the code and this is not needed now. thanks

src/lib/commandBuilder.ts
dennisseah: refactor the code and this is not needed now. thanks

src/lib/commandBuilder.ts
dennisseah: accept thanks

src/lib/validator.ts

src/commands/infra/scaffold.ts
```diff
@@ -5,9 +5,86 @@ import yaml from "js-yaml";
...
+      removeTemplateFiles(opts.name);
+      exitFn(0);
+    } catch (err) {
+      logger.error("Error occurred while generating scaffold");
+      logger.error(err);
+      exitFn(1);
+    }
+  }
+};
+
```
yradsmikham: Somewhere in this refactoring is breaking the functionality of `spk infra scaffold`. After testing the code, I am unable to generate a ` definition.yaml` file. I tried debugging, but am unable to find the root cause. Will need to dive into it a little deeper...

But you if you run `spk infra scaffold` against the spk master branch, and one against your forked branch, you'll see that in the refactored code, scaffolding fails to generate the definition and remove the templated terraform files.

SPK master branch:
Screen Shot 2020-01-22 at 8 57 30 AM

Your SPK master branch:
Screen Shot 2020-01-22 at 8 58 33 AM

Are you able to reproduce the same behavior on your side?

yradsmikham (changes requested):

src/commands/infra/scaffold.ts
dennisseah: chat with Yvonne and figure out the problem.

yradsmikham:
LGTM

yradsmikham (approved):

evanlouie (approved):

PR #194, opened 2020-01-21
deployment onboard integration test

edaena:


tests/functions.sh
```diff
@@ -1,6 +1,6 @@
 print_array () {
     #usage: print_array "${my_array[@]}"
-    arr=("$@")
+    arr=( "$@" )
```
edaena: remove space

andrebriggs: @edaena are you planning to revert this change?

andrebriggs (approved):

samiyaakhtar (approved):

tests/functions.sh
edaena: @andrebriggs reverted the change and also reverted the requirement for subscription id, it will just use the default one that's on `az login`

mtarng (approved):

PR #197, opened 2020-01-22
[REFACTOR] commands/hld/init.ts and add tests

dennisseah:
1. move the command decoration out to a JSON file
2. standardized on command decorator function name, `initCommandDecorator` -> `commandDecorator`
3. added support for having default values in command builder
4. added tests to boost the coverage from 59% to 89%
5. create helper functions in tests so test code is more manage-able


evanlouie (approved):
lgtm 👍

andrebriggs (approved):

PR #201, opened 2020-01-23
Specify Traefik2 IngressRoute Service Port

### LINKED ISSUE #818: Add the ability to specify ports for traefik ingress routes
**As a:**

SPK User

**I want:**

To be able to specify the port for services in my bedrock.yaml for backing k8s services

**So that:**

When the HLD lifecycle pipeline creates my traefik ingress route, it is able to use the service specified in my bedrock.yaml file to scaffold the backing service port

bedrock.yaml file schema to be updated to accept a `port` argument for services configuration.
hld lifecycle pipeline to use the port specified on services in bedrock.yaml to scaffold traefik ingress routes

**Acceptance Criteria:**

- [ ] I can specify --k8s-port when creating services with `spk` to configure a traefik ingress route.
- [ ] HLD lifecycle pipeline to scaffold traefik ingress routes with the port specified from bedrock.yaml 

Not supporting ingress route configuration and making the presumption that the backing k8s port is always the same

**Additional context:**

Traefik ingress routes must be configured by specifying a port for the backing k8s service. When creating/scaffolding the HLD, the port for the backing service is unknown (it usually provided as configuration on helm templates).

Yes

evanlouie:
Added the ability to specify a `--k8s-service-port` to `spk service create`.
This port will now be tracked in the users `bedrock.yaml` file and be injected
to the Traefik2 IngressRoutes created via `spk hld reconcile`.

closes https://github.com/microsoft/bedrock/issues/818

### DIFF
--- a/src/commands/hld/reconcile.ts
+++ b/src/commands/hld/reconcile.ts
@@ -264,13 +264,17 @@ const createIngressRouteForRing = (
     staticComponentPathInRing,
     "ingress-route.yaml"
   );
-  // TODO: figure out a way to grab the port from _somewhere_; store in bedrock.yaml?
-  const ingressRoute = TraefikIngressRoute(serviceName, ring, 8000, {
-    middlewares: [
-      middlewares.metadata.name,
-      ...(serviceConfig.middlewares ?? [])
-    ]
-  });
+  const ingressRoute = TraefikIngressRoute(
+    serviceName,
+    ring,
+    serviceConfig.k8sServicePort,
+    {
+      middlewares: [
+        middlewares.metadata.name,
+        ...(serviceConfig.middlewares ?? [])
+      ]
+    }
+  );
 
   const routeYaml = yaml.safeDump(ingressRoute, {
     lineWidth: Number.MAX_SAFE_INTEGER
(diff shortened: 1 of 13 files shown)


src/commands/hld/reconcile.ts
```diff
@@ -264,8 +264,8 @@ const createIngressRouteForRing = (
     staticComponentPathInRing,
     "ingress-route.yaml"
   );
-  // TODO: figure out a way to grab the port from _somewhere_; store in bedrock.yaml?
-  const ingressRoute = TraefikIngressRoute(serviceName, ring, 8000, {
+  const k8sServicePort = serviceConfig.k8sServicePort ?? 80;
```
evanlouie: Note. Because the `k8sServicePort` for `IBedrockServiceConfig` is possibly undefined. The default behaviour is to use port 80.
The `spk service create` command also defaults to creating the service with port 80, but wasn't sure if making the `k8sServicePort` field in `bedrock.yaml` required was best -- thoughts?

mtarng (approved):

bnookala (approved):

PR #205, opened 2020-01-24
Use display name for creating service components and ingress routes

### LINKED ISSUE #898: ingress routes do not use display name
**Describe the bug:**

Ingressroutes being written improperly - not using displayName attribute in service object

**To Reproduce:**

bedrock.yaml: 
```
rings:
  master:
    isDefault: true
services:
  ./:
    displayName: fabrikam-frontend-service
    helm:
      chart:
        branch: master
        git: 'https://dev.azure.com/foo/bar/baz'
        path: chart
    middlewares:
      - ''
```

will produce:

```
kind: IngressRoute
metadata:
  name: .-master
spec:
  routes:
    - kind: Rule
      match: 'PathPrefix(`/.`) && Headers(`Ring`, `master`)'
      services:
        - name: .-master
          port: 8000
```

where the name of the ingress route, the path prefix, and the service name are all wrong (using the path to the service, instead of the display name0

**Expected behavior:**

we should be outputting the correct service name, correct ingress route name using the ingress display name, and correct path prefix

**Screenshots:**

**Additional context:**

presently we're not using `displayName` at all

no

Related to #810

bnookala:
Fixes https://github.com/microsoft/bedrock/issues/898

### DIFF
--- a/src/commands/hld/reconcile.ts
+++ b/src/commands/hld/reconcile.ts
@@ -151,15 +151,24 @@ export const reconcileHld = async (
   for (const [serviceRelPath, serviceConfig] of Object.entries(
     managedServices
   )) {
-    const pathBase = path.basename(serviceRelPath);
-    const serviceName = pathBase;
-    logger.info(`Reconciling service: ${pathBase}`);
+    const serviceName =
+      serviceConfig.displayName || path.basename(serviceRelPath);
+
+    // No name, cannot generate proper routes and middlewares
+    if (serviceName === "." || !serviceName) {
+      logger.warn(
+        "Service displayName not provided or service path is `.` - not reconciling service"
+      );
+      continue;
+    }
+
+    logger.info(`Reconciling service: ${serviceName}`);
 
     // Utilizes fab add, which is idempotent.
     await dependencies.createServiceComponent(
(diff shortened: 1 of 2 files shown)


src/commands/hld/reconcile.ts
```diff
@@ -151,15 +151,24 @@ export const reconcileHld = async (
...
   )) {
-    const pathBase = path.basename(serviceRelPath);
-    const serviceName = pathBase;
-    logger.info(`Reconciling service: ${pathBase}`);
+    const serviceName =
+      serviceConfig.displayName || path.basename(serviceRelPath);
+
+    // No name, cannot generate proper routes and middlewares
+    if (serviceName === "." || !serviceName) {
+      logger.warn(
```
evanlouie, line 159: might want to throw an error and exit instead

evanlouie (approved):
lgtm

src/commands/hld/reconcile.ts
bnookala: might be a larger design choice as there's a few places where we simply just `continue` instead of failing fast - perhaps we can revisit this

mtarng (approved):

PR #209, opened 2020-01-27
[REFACTOR] Refactor generate ts

dennisseah:
1. move the command decorate to a JSON File
2. add tests due to refactoring
3. code simplification
4. exit code 0 when there are no errors
5. use enum for different configuration
6. make sure the functions do not return Promise if not necessary


src/commands/infra/generate.test.ts
```diff
@@ -31,206 +142,225 @@ describe("test dirIteration", () => {
...
-      "A",
-      "https://github.com/yradsmikham/spk-source",
-      "cluster/environments/azure-single-keyvault",
-      "v0.0.2"
-    ];
-    let sourceConfiguration = await validateDefinition(
+  test("definition.yaml of leaf override parent's variable", async () => {
+    const mockParentPath = "src/commands/infra/mocks/discovery-service";
+    const mockProjectPath = "src/commands/infra/mocks/discovery-service/west";
+    const expectedSourceWest = {
```
andrebriggs, line 258: More readable 👍🏾

src/commands/infra/generate.test.ts
```diff
@@ -31,206 +142,225 @@ describe("test dirIteration", () => {
...
-    sourceConfiguration = await validateDefinition(
+    expect(sourceData).toEqual(expectedSourceWest);
+    await generateConfig(
+      mockParentPath,
+      mockProjectPath,
+      sourceConfiguration,
+      sourceData
+    );
+  });
+  test("without parent's definition.xml", async () => {
```
andrebriggs: @dennisseah you mean yaml instead of xml?

src/commands/infra/generate.test.ts
```diff
@@ -31,206 +142,225 @@ describe("test dirIteration", () => {
...
       mockParentPath,
-      mockParentPath
+      mockProjectPath,
+      sourceConfiguration,
+      sourceData
     );
-
-    expect(sourceConfiguration).toEqual("");
+  });
+  test("without parent's and project's definition.xml", async () => {
```
andrebriggs: Same here

andrebriggs (changes requested):
@dennisseah looks good. A small nitpick. I assume this was integration tested?

cc @yradsmikham @NathanielRose

src/commands/infra/generate.test.ts
dennisseah: yes. thanks

src/commands/infra/generate.test.ts
dennisseah: yes. thanks

src/commands/infra/generate.ts
```diff
@@ -180,61 +195,63 @@ export const validateTemplateSources = async (
...
+
+    await simpleGit(sourcePath).checkout(version);
   } catch (err) {
-    logger.error(
-      `There was an error checking the remote source repository: ${err}`
-    );
-    return false;
+    logger.error(err);
+    // TOFIX: this error should be rethrown
+    // and if we do, infra-validations.sh will fail
```
yradsmikham, line 252: What do you mean by this error should be "rethrown"?

src/commands/infra/generate.ts
```diff
@@ -434,6 +437,7 @@ export const generateConfig = async (
         );
       }
     }
+    // TODO: handle definitionConfig === DefinitionYAMLExistence.LEAF_ONLY
```
yradsmikham, line 439: @NathanielRose @andrebriggs 

Revisiting this discussion... do we want to support scenarios where there is no `definition.yaml` and only leaf definitions?

src/commands/infra/generate.ts
```diff
@@ -180,61 +195,63 @@ export const validateTemplateSources = async (
...
-      }
-      // Checkout tagged version
-      logger.info(`Checking out template version: ${version}`);
-      await simpleGit(sourcePath).checkout(version);
+      await gitClone(source, sourcePath);
     }
+    // Checkout tagged version
+    logger.info(`Checking out template version: ${version}`);
+
+    await simpleGit(sourcePath).checkout(version);
```
yradsmikham, line 248: Something in this step is returning an empty message:
Screen Shot 2020-01-28 at 9 28 17 AM

This happens when I run in a hierarchy where there is no parent definition.yaml and a there exists a leaf definition.yaml.

yradsmikham (changes requested):

src/commands/infra/generate.ts
dennisseah: if I do `throw err` so that the caller function can handle the exception and quit. then `infra-validations.sh` will not run to the end. This is what Nate mentioned about issue https://github.com/microsoft/bedrock/issues/893. We can `throw err` once this issue is address and `infra-validations.sh` can complete successfully.

src/commands/infra/generate.ts
dennisseah: this empty info: liner will go away after this PR is merged https://github.com/CatalystCode/spk/pull/214.
The case where parent definition.yaml is missing not not handled. You, Nate and Andre (who are the domain experts) can to decide what to do. And it can be handle in a different PR (not making this PR too big).

src/commands/infra/generate.ts
dennisseah: Once decision is made, this can be handled in a separate PR with a new user story. and i can work on it if needed

yradsmikham: Had a small conversation with @NathanielRose, and the consensus right now is opt out of having a "configuration C" and throw an error for the case of a missing parent `definition.yaml`. The reason is because `spk infra` is really designed around managing multi-cluster scenarios for Bedrock. Having individual infrastructure configuration would imply a _non_-multi-cluster setup, and in which case, the user should just handle the operations manually and not rely on `spk infra generate`.

yradsmikham: If it's not too much to ask for, could you include this error handling in the code? :)

andrebriggs: @yradsmikham This sounds like a separate issue. @dennisseah @yradsmikham Let's a new issue for this bug/feature and address it in a new PR. Let's keep PR sizes small when possible

src/commands/infra/generate.ts
yradsmikham: Seems like that PR is merged, and I am still seeing that empty line.

src/commands/infra/generate.ts
yradsmikham: @dennisseah @andrebriggs sure, I can create task for this, and work on it.

andrebriggs: @yradsmikham please share the new issue to groomed once you've created it

yradsmikham: @andrebriggs https://github.com/microsoft/bedrock/issues/918

src/commands/infra/generate.ts
dennisseah: this is going to change again. see PR: https://github.com/CatalystCode/spk/pull/221

andrebriggs: Thanks for the update @dennisseah

andrebriggs (approved):

src/commands/infra/generate.ts
yradsmikham: I see that is it is resolved now by appending `--end log: 0--` at the end of the log. Tested.

yradsmikham (approved):

NathanielRose (approved):

PR #218, opened 2020-01-27
Setup pipelines in introspection tests file

edaena:
Service introspection provisioning
- Create new pipelines for introspection `fabrikamintro2019-lifecycle` and `fabrikam-intro-hld-to-fabrikam-manifests`
- Setup variable group for introspection
- Remove service introspection logic from `validations.sh`


andrebriggs (approved):
Looks good. We can work offline to make sure that any cloud resources we need in the integration test work for the automated test run

samiyaakhtar (approved):

PR #215, opened 2020-01-27
[REFACTOR] project pipeline command

dennisseah:
1. Create a set of bedrockYaml helper functions so that 
   we have a common place for `bedrock.yaml` related functions
   a. e.g. isBedrockFileExists function is in create-variable-group.ts
      and it is consumed by create-variable-group-test.ts.
   b. create a dummy `bedrock.yaml` in a random temporay directory is 
      also a common used one
   c. reading `bedrock.yaml` as a JSON object is another one
   d. create a constant `YAML_NAME` for `bedrock.yaml`
   e. have a `addNewService` function
   will expand to have more features like write `bedrock.yaml`, etc.

2. Remove the validateRequiredArguments function in create-variable-group.ts
   the functionality is handled by `CommandBuilder.validateForRequiredValues`
   function

3. simplify pipeline.test.ts as mock values and objects are repeated many times

4. pipeline.ts
   a. move command decoration out to a JSON file
   b. declare an interface for the expected values from commandline
   c. replace `isValidConfig` function with `CommandBuilder.validateForRequiredValues`

5. miscellaneous changes.

will fix the status code issue once this PR is merged https://github.com/CatalystCode/spk/pull/214

PR #220, opened 2020-01-28
Service pipelines using variable groups from bedrock.yaml

### LINKED ISSUE #887: [spk] Service pipelines should only use project Variable group
**As a:** SPK user and developer

**I want:** service pipelines to only utilize variable groups for variables
instead of those baked into the yaml

**So that:** the utilization of variable groups across lifecycle and service
pipelines is standardized across the project

- [x] Refactor `spk service create` to output azure-piplines.yaml that consume
      the variable groups created in `spk project cvg`
- [x] Refactor `spk service create` to remove the `--variable-group-name <variable-group-name` flag, as spk should only use variable groups defined under the project's `bedrock.yaml
- [x] Check if user executed `spk project init` and `spk project cvg` commands before `spk service create` command for dependencies

**Acceptance Criteria:**

refer to task list above

**Additional context:**

[Project commands are now expected to run in order](https://github.com/microsoft/bedrock/issues/886)

**Does this require updates to documentation?:** yes

sarath-p:
- Closes https://github.com/microsoft/bedrock/issues/887
- Checks if user ran `spk project init` and `spk project cvg` commands before this `spk service create`


dennisseah:
We shall have conflict with other PR that I have https://github.com/CatalystCode/spk/pull/215
I have moved isBedrockFileExist to a new helper file where it contains all the bedrock related functions.

sarath-p:
> We shall have conflict with other PR that I have #215
> I have moved isBedrockFileExist to a new helper file where it contains all the bedrock related functions.

The function code also changed. We can merge through master branch.

src/config.ts
```diff
@@ -294,3 +294,31 @@ export const saveConfiguration = async (
...
+  try {
+    bedrockFile = await BedrockAsync(absProjectPath);
+  } catch (error) {
+    logger.error(error);
+    return [false, undefined];
+  }
+
+  logger.debug(`variableGroups length: ${bedrockFile?.variableGroups?.length}`);
+  logger.verbose(`bedrockFile: \n ${bedrockFile}`);
+  return [true, bedrockFile?.variableGroups?.length];
```
dennisseah: you are overloading this function. I hope that you can return `true` or `false` here. and then do variableGroups length check in the caller function. When we overload function, you can see that you needed to return things like undefined. thanks.

sarath-p: It is a deep check i.e. not only a file exist, but need to make sure variable groups exist in the file. In this scenario, the caller does't need the file contents and true/false is sufficient. Perhaps we can change the second one to `boolean` and remove the `undefined`.  Let me know your thoughts on this.

dennisseah: I am misled by the function name, isBedrockFileValid. As the function name suggest, I thought that we are checking if bedrock file exist or not :-) I think we can do this.
```
interface IBedrockFileStat {
  exist: boolean;
  hasVariableGroups: boolean;
}

export const bedrockFileStat = async (
  rootProjectPath: string
): Promise<IBedrockFileStat> => {
  if (!rootProjectPath) {
    throw new Error("Project root path is not valid");
  }

  const absProjectPath = path.resolve(rootProjectPath);

  let bedrockFile: IBedrockFile | undefined;

  try {
    const bedrockFile = await BedrockAsync(absProjectPath);
    return {
      exist: true,
      hasVariableGroups: bedrockFile?.variableGroups?.length
    }
  } catch (error) {
    logger.error(error);
    return {
      exist: false,
    }
  }
}
```

this enables other team members to reuse and enhance this function get more stats from bedrock file.
and we do not return array, we return a well defined interface.

what do you think?

sarath-p: Yes, the name was misleading. I thought about it and am going to refactor it to check whether dependency commands were executed.  One of the goal it to centralize the error messages too so they won't be repeated.

sarath-p: Discussed with @mtarng and the decided to keep it as `bedrock file info` and use it to validate the project commands order.

dennisseah: sorry, I do not understand `keep it as bedrock file info`. Do you mind elaborate?

sarath-p: I was thinking to implement command sequence validator using this logic, but decided to keep the method name as bedrock file that may be used for other checks in the future.

dennisseah: ok

dennisseah (approved):

src/commands/project/create-variable-group.ts
```diff
@@ -76,72 +82,80 @@ export const execute = async (
...
-        logger.info(
-          "Successfully created a variable group in Azure DevOps project!"
-        );
-        await exitFn(0);
-      }
-    } catch (err) {
-      logger.error(`Error occurred while creating variable group`);
-      logger.error(err);
-      await exitFn(1);
+    const bedrockFileIno = await bedrockFileInfo(projectPath);
```
mtarng: Typo?
>bedrockFileIno

or bedrockFileInfo

sarath-p: yes, let me fix it

mtarng (approved):

PR #221, opened 2020-01-28
revert the command exit function

dennisseah:


src/lib/commandBuilder.ts
```diff
@@ -118,10 +118,9 @@ export const exit = (
   statusCode: number
 ): Promise<void> => {
   return new Promise(resolve => {
-    log.end();
-    setTimeout(() => {
+    log.info("--end log--", () => {
```
andrebriggs: @dennisseah so `--end log--` will be at the end of each call in the log regardless of success?

dennisseah: > @dennisseah so `--end log--` will be at the end of each call in the log regardless of success?

correct

andrebriggs (approved):

samiyaakhtar (approved):

PR #222, opened 2020-01-29
Adding configs for ingress routes for pathprefix, version, backend.

### LINKED ISSUE #832: Scaffold ingress routes with major version
**As a:**

SPK user

**I want:**

bedrock.yaml to offer overrides for ingress route path prefixes and major version

**So that:**

The scaffolded ingress route resembles the routes advised by the MS Rest guidelines https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md

For a bedrock.yaml that resembles

```
rings:
  dev:
    isDefault: true
  prod:
services:
  homepage:
    port: 8001
    pathPrefix: homepage
    version: v1
    backend: k8s-backend-service
```

I want a scaffolded ingress route for all rings that produces a path prefix of:

/v1/homepage 

instead of using just the service name as the route 

**Acceptance Criteria:**

- A bedrock.yaml file that resembles:

```
rings:
  dev:
    isDefault: true
  prod:
services:
  homepage:
    port: 8001
    pathPrefix: homepage
    version: v1
    backend: k8s-backend-service
```

- [ ] should output a traefik ingress route that resembles:

```
kind: IngressRoute
metadata:
  name: homepage-dev
spec:
  routes:
    - kind: Rule
      match: 'PathPrefix(`/v1/homepage`) && Headers(`Ring`, `dev`)'
      services:
        - name: k8s-backend-service-dev
          port: 8001
```

- [ ] a test to cover the correct-ness of the scenario

None

**Additional context:**

We require more flexibility with the ingress routes that are being scaffolded

Yes

mtarng:
Closes https://github.com/microsoft/bedrock/issues/832

### DIFF
--- a/docs/service-management.md
+++ b/docs/service-management.md
@@ -50,9 +50,9 @@ Options:
   --git-push                                                  SPK CLI will try to commit and push these changes to a new origin/branch named after the service. (default: false)
   --middlewares <comma-delimitated-list-of-middleware-names>  Traefik2 middlewares you wish to to be injected into your Traefik2 IngressRoutes (default: "")
   --k8s-service-port <port>                                   Kubernetes service port which this service is exposed with; will be used to configure Traefik2 IngressRoutes (default: "80")
+  --k8s-backend <backend>                                     Kubernetes service name; will be used to configure Traefik2 IngressRoutes (default: "")
   --path-prefix <path-prefix>                                 The path prefix for ingress route; will be used to configure Traefik2 IngressRoutes. If omitted, then the service name will used. (default: "")
-  --path-prefix-version <path-prefix-version>                 Version to be used in the path prefix; will be used to configure Traefik2 IngressRoutes. ie. 'v1' will result in a path prefix of '/v1/servicename (default: "")
-  --backend <backend>                                         Kubernetes service name; will be used to configure Traefik2 IngressRoutes (default: "")
+  --path-prefix-major-version <path-prefix-major-version>     Version to be used in the path prefix; will be used to configure Traefik2 IngressRoutes. ie. 'v1' will result in a path prefix of '/v1/servicename (default: "")
   -h, --help                                                  output usage information
 ```
 
@@ -63,7 +63,7 @@ Options:
 - `--middlewares`, `--k8s-service-port`, `--path-prefix`,
   `--path-prefix-version`, and `--backend` are all used to configure the
   generated Traefik2 IngressRoutes. ie.
-  `spk service create my-example-documents-service --middlewares middlewareA --k8s-service-port 3001 --path-prefix documents --path-prefix-version v2 --backend docs-service`
+  `spk service create my-example-documents-service --middlewares middlewareA --k8s-service-port 3001 --k8s-backend docs-service --path-prefix documents --path-prefix-major-version v2`
   will result in an IngressRoute that looks like:
   ```
   apiVersion: traefik.containo.us/v1alpha1
--- a/docs/service-management.md
+++ b/docs/service-management.md
@@ -50,6 +50,9 @@ Options:
   --git-push                                                  SPK CLI will try to commit and push these changes to a new origin/branch named after the service. (default: false)
   --middlewares <comma-delimitated-list-of-middleware-names>  Traefik2 middlewares you wish to to be injected into your Traefik2 IngressRoutes (default: "")
   --k8s-service-port <port>                                   Kubernetes service port which this service is exposed with; will be used to configure Traefik2 IngressRoutes (default: "80")
+  --path-prefix <path-prefix>                                 The path prefix for ingress route; will be used to configure Traefik2 IngressRoutes. If omitted, then the service name will used. (default: "")
+  --version <version>                                         Version to be used in the path prefix; will be used to configure Traefik2 IngressRoutes. ie. 'v1' will result in a path prefix of '/v1/servicename (default: "")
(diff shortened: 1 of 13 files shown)


andrebriggs:
Looks good so far! @mtarng an example usage of these new parameters in documentation would be great. Also can we start adding folks from outside the "sub-group" as reviewers?

andrebriggs (changes requested):
Docs

src/commands/project/init.ts
```diff
@@ -157,14 +156,12 @@ const generateMaintainersFile = async (
...
   projectPath: string,
-  packagePaths: string[],
   defaultRings: string[] = []
 ) => {
   const absProjectPath = path.resolve(projectPath);
-  const absPackagePaths = packagePaths.map(p => path.resolve(p));
   logger.info(`Generating bedrock.yaml file in ${absProjectPath}`);
 
-  const base: IBedrockFile = {
+  const baseBedrockFile: IBedrockFile = {
```
dennisseah, line 164: 👍

src/commands/service/create.ts
```diff
@@ -90,6 +90,21 @@ export const createCommandDecorator = (command: commander.Command): void => {
       "Kubernetes service port which this service is exposed with; will be used to configure Traefik2 IngressRoutes",
       "80"
     )
+    .option(
+      "--path-prefix <path-prefix>",
+      "The path prefix for ingress route; will be used to configure Traefik2 IngressRoutes. If omitted, then the service name will used.",
+      ""
+    )
+    .option(
+      "--version <version>",
```
dennisseah: can we have `--path-prefix-version` to clearly denote what this version is for?

dennisseah:
FYI, `addNewServiceToBedrockFile` function is moved to a new file in my PR. https://github.com/CatalystCode/spk/pull/215/files#diff-3518fb7be6e773dea259d8783503a1cfR75

docs/service-management.md
```diff
@@ -51,6 +51,9 @@ Options:
   --variable-group-name <variable-group-name>                 The Azure DevOps Variable Group.
   --middlewares <comma-delimitated-list-of-middleware-names>  Traefik2 middlewares you wish to to be injected into your Traefik2 IngressRoutes (default: "")
   --k8s-service-port <port>                                   Kubernetes service port which this service is exposed with; will be used to configure Traefik2 IngressRoutes (default: "80")
+  --path-prefix <path-prefix>                                 The path prefix for ingress route; will be used to configure Traefik2 IngressRoutes. If omitted, then the service name will used. (default: "")
```
bnookala, line 54: if the default is the service name, then default: "" seems false-y

docs/service-management.md
```diff
@@ -51,6 +51,9 @@ Options:
...
+  --version <version>                                         Version to be used in the path prefix; will be used to configure Traefik2 IngressRoutes. ie. 'v1' will result in a path prefix of '/v1/servicename (default: "")
```
bnookala: can we re-name this to `--major-version`?

src/commands/service/create.ts
```diff
@@ -169,18 +193,24 @@ export const createCommandDecorator = (command: commander.Command): void => {
...
- * @param helmConfigGit Helm git
- * @param helmConfigPath Helm config path
- * @param serviceName Service name
- * @param packagesDir Packages directory
- * @param maintainerName Name of maintainer
- * @param maintainerEmail Email of maintainer
- * @param middlewares comma-delimitated list of Traefik2 middlewares
- * @param gitPush Push to git
- * @param variableGroupName Variable group name
+ *
```
bnookala, line 204: perhaps this is something we can create an interface for - since it's getting quite unwieldy

andrebriggs (approved):

src/commands/service/create.ts
```diff
@@ -89,6 +89,21 @@ export const createCommandDecorator = (command: commander.Command): void => {
       "Kubernetes service port which this service is exposed with; will be used to configure Traefik2 IngressRoutes",
       "80"
     )
+    .option(
+      "--path-prefix <path-prefix>",
+      "The path prefix for ingress route; will be used to configure Traefik2 IngressRoutes. If omitted, then the service name will used.",
+      ""
+    )
+    .option(
+      "--path-prefix-version <path-prefix-version>",
```
bnookala: --path-prefix-major-version

src/commands/service/create.ts
```diff
@@ -89,6 +89,21 @@ export const createCommandDecorator = (command: commander.Command): void => {
...
+      "Version to be used in the path prefix; will be used to configure Traefik2 IngressRoutes. ie. 'v1' will result in a path prefix of '/v1/servicename",
+      ""
+    )
+    .option(
+      "--backend <backend>",
```
bnookala: maybe something like --k8s-backend

bnookala (changes requested):

docs/service-management.md
mtarng: The problem is that this documentation is generated by Commander. I don't think we can change what goes in the default column, otherwise it'll have a downstream impact on all path prefixes

src/commands/service/create.ts
evanlouie: we currently have `--k8s-service-port` to specify the port for it. perhaps `--k8s-service-name`

src/commands/service/create.ts
mtarng: Yeah, I agree. We should do that for each command, but I'd like to keep that outside the scope of this PR

bnookala (approved):

PR #225, opened 2020-01-29
[REFACTOR] Refactor scaffold.ts, add more tests and fix 922

dennisseah:
1. move command decorator to a json file
2. add more tests to boast coverage to 80 -90% -ish
3. declare constants for frequently used strings
4. change ICommandOptions to be more specified on the property keys
5. Fix issue https://github.com/microsoft/bedrock/issues/922
6. add await to command exit function
7. throw exception in validating function so that caller will not miss them
8. remove unused function, renameTfvars
9. make some functions non promisified
10. avoid having many parameters in functions


src/commands/infra/scaffold.ts
```diff
@@ -96,89 +100,49 @@ export const execute = async (
...
-      "-v, --version <repository version>",
-      "Version or tag for the repository so a fixed version is referenced"
-    )
-    .option(
-      "-t, --template <path to variables.tf> ",
-      "Location of the variables.tf for the terraform deployment"
-    )
-    .action(async (opts: ICommandOptions) => {
-      const config = Config();
-      await execute(config, opts, process.exit);
```
yradsmikham, line 119: Is there a reason why we abstract the command decorator in a `json` file?  Is this a best practice? It just seems like this is just another way of describing the scaffold command, but not necessarily a _better_ way. 

My only concern with doing this is, now we are introducing inconsistencies across `spk`. Do we plan on making this change for `spk infra generate`? or even other `spk` commands as well?

yradsmikham (changes requested):

src/commands/infra/scaffold.ts
dennisseah: the plan to move declaration like this out of code. I already did for generate.ts and will do for the rest of the command files as part of https://github.com/microsoft/bedrock/issues/885 effort

src/commands/infra/scaffold.test.ts
```diff
@@ -1,14 +1,29 @@
 jest.mock("./generate");
```
NathanielRose, line 2: This test has a new dependency on mock generate and scaffold folders?

src/commands/infra/scaffold.ts
yradsmikham: Okay, thanks for the clarification!

yradsmikham (approved):

src/commands/infra/scaffold.ts
dennisseah: Thanks Yvonne, the epic is not clear until Andre pointed out to me that I have added more information. Sorry for the confusion.

andrebriggs (approved):

dennisseah:
fixed https://github.com/microsoft/bedrock/issues/922

PR #227, opened 2020-01-29
Fixed typo, clarified comments in spk-config.yaml

xtophs:
Fixed typo in project-service-management-guide.md
Removed PAT from spk-config.yaml
Removed default values that could cause problems and confusion for users from spk-config.yaml.
Added clarifying comments to spk-config.yaml


spk-config.yaml
```diff
@@ -1,11 +1,11 @@
...
 azure_devops:
-  access_token: "hpe3a9oiswgcodtfdpzfiek3saxbrh5if1fp673xihgc5ap467a" # This is your Personal Access Token with permission to modify and access this private repo. Leave this empty if project is public
-  hld_repository: "https://dev.azure.com/bhnook/fabrikam/_git/hld" # Repository URL for your Bedrock HLDs
-  manifest_repository: "https://dev.azure.com/bhnook/fabrikam/_git/materialized" # Repository URL that is configured for flux. This holds the kubernetes manifests that is generated by fabrikate.
-  infra_repository: "https://dev.azure.com/naros32/bedrock_testing_phase2/_git/spk-testing-private-azdo" # Repository URL that contains your terraform templates to be sed for scaffolding and generating infrastructure deployment templates.
-  org: "bhnook" # Your AzDo Org
-  project: "fabrikam" # Your AzDo project
+  access_token: "" # This is a Personal Access Token with permission to modify and access the HLD, manifest and infra repos. Leave this empty if project is public
+  hld_repository: "" # Repository URL for your Bedrock HLDs
+  manifest_repository: "" # Repository URL for yor materialized manifests generated by fabrikate.
```
bnookala: your

bnookala (commented):
one nitpick, but would like us to consider if we want to remove the repository URLs and org/project names, as they're a "hint" to the user on what the format should be

spk-config.yaml
```diff
@@ -1,11 +1,11 @@
 # Azure DevOps Configuration
 azure_devops:
-  access_token: "hpe3a9oiswgcodtfdpzfiek3saxbrh5if1fp673xihgc5ap467a" # This is your Personal Access Token with permission to modify and access this private repo. Leave this empty if project is public
-  hld_repository: "https://dev.azure.com/bhnook/fabrikam/_git/hld" # Repository URL for your Bedrock HLDs
-  manifest_repository: "https://dev.azure.com/bhnook/fabrikam/_git/materialized" # Repository URL that is configured for flux. This holds the kubernetes manifests that is generated by fabrikate.
-  infra_repository: "https://dev.azure.com/naros32/bedrock_testing_phase2/_git/spk-testing-private-azdo" # Repository URL that contains your terraform templates to be sed for scaffolding and generating infrastructure deployment templates.
-  org: "bhnook" # Your AzDo Org
-  project: "fabrikam" # Your AzDo project
+  access_token: "" # This is a Personal Access Token with permission to modify and access the HLD, manifest and infra repos. Leave this empty if project is public
```
mtarng: For specifics on which permissions this PAT needs, they are listed in this doc: https://github.com/CatalystCode/spk/blob/master/docs/project-service-management-guide.md#generating-personal-access-token

Do you think we should also include those here?

spk-config.yaml
mtarng: Typo
>yor

xtophs: The URLs with real GitHub users that could be interpreted as repos that should already exist were confusing. Generic URLs that show the pattern but make it clear to the user that those are project specific could be helpful.

xtophs: fixed

spk-config.yaml
xtophs: Added the link to the doc.

bnookala (approved):
Thanks for generalizing the URLs!

spk-config.yaml
xtophs: added dummy URLs, fixed type

mtarng (approved):

PR #231, opened 2020-01-30
Fixing project init test that left files behind

### LINKED ISSUE #926: [spk] spk yarn tests produces lasting artifacts
**Describe the bug:**
Running yarn test creates and leaves three files in the working directory:
`bedrock.yaml`
`hld-lifecycle.yaml`
`maintainers.yaml`

**To Reproduce:**
- Download the latest spk code
- Run `spk yarn`
- Run `spk yarn test`

**Expected behavior:**
`spk yarn test` should succeed and create no new files.

**Screenshots:**

**Additional context:**

mtarng:
Closes https://github.com/microsoft/bedrock/issues/926


evanlouie (approved):

bnookala (approved):
i _think_ the pattern we should adopt is to pass in the writeFile function as a dependency to whichever function needs it (and use a mock jest.fn to confirm its being called at all), but we can ignore it for now

PR #233, opened 2020-01-30
Nathaniel rose patch 1

NathanielRose:
Compile error added in revert for variable group.

***Update*** Separated variable group per each test to pass: Build Successful for the patch-https://dev.azure.com/epicstuff/bedrock/_build/results?buildId=10569&view=logs&j=c43442fc-b3ed-5a4c-609f-f3931a1f4b55&t=c43442fc-b3ed-5a4c-609f-f3931a1f4b55

-  Smoke Test Fails on current VG: https://dev.azure.com/epicstuff/bedrock/_build/results?buildId=10541&view=logs&j=c43442fc-b3ed-5a4c-609f-f3931a1f4b55&t=7da6109c-585e-54f6-c567-ddc1bd543ed7
- Passes on Test VG: https://dev.azure.com/epicstuff/bedrock/_build/results?buildId=10546&view=results

### DIFF
--- a/smoke-test-pipeline.yml
+++ b/smoke-test-pipeline.yml
@@ -8,7 +8,7 @@ schedules:
   always: true
 
 variables:
-- group: 'spk-vg-testing'
+- group: 'spk-vg'
 
 stages:
 - stage: integration_tests
--- a/smoke-test-pipeline.yml
+++ b/smoke-test-pipeline.yml
@@ -103,7 +103,7 @@ stages:
         itemPattern: "spk*"
         artifactName: 'spk_linux_node_12'
         targetPath: '$(System.DefaultWorkingDirectory)'
-    - bash: |in 
+    - bash: | 
         shopt -s expand_aliases
         echo "Current Dir is $(pwd)"
         chmod +x spk-linux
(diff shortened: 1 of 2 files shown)


smoke-test-pipeline.yml
```diff
@@ -8,7 +8,7 @@ schedules:
   always: true
 
 variables:
-- group: 'spk-vg'
+- group: 'spk-vg-testing'
```
andrebriggs: @NathanielRose You're proposing to change the environment (via new variable group) in order to get infrastructure tests to pass but we still don't know the root cause of why we have to change the environment?

It would be nice to know exactly what the difference is here so we could add it to our documentation around integration test set up. 

Right now it seems like we are make blind changes and things _just happen_ to work

smoke-test-pipeline.yml
```diff
@@ -103,7 +103,7 @@ stages:
         itemPattern: "spk*"
         artifactName: 'spk_linux_node_12'
         targetPath: '$(System.DefaultWorkingDirectory)'
-    - bash: |in
```
andrebriggs, line 106: How did this work?

NathanielRose: This is a compile error that was added when the variable group was switched

smoke-test-pipeline.yml
NathanielRose: I can revert the test for infra back to v1 until I have a solution to why the SP in the prod variable group is throwing the permissions error?

yradsmikham (approved):

andrebriggs (approved):

PR #234, opened 2020-01-30
[REFACTOR] Service create command

dennisseah:

2. call common exit function
3. added test to boost coverage to 90%-ish
4. added better validation for port
5. have an interface for command values to avoid passing many values to functions


mtarng (approved):

src/commands/service/create.ts
```diff
@@ -10,328 +10,144 @@ import {
...
+    maintainerName: opts.maintainerName,
+    middlewares: opts.middlewares,
+    middlewaresArray,
+    packagesDir: opts.packagesDir,
+    pathPrefix: opts.pathPrefix,
+    pathPrefixMajorVersion: opts.pathPrefixMajorVersion,
+    variableGroups
+  };
+
+  // values need not be validated (that's do not need
```
mtarng: Wording here is odd.

> Values do not need to be validated as they are mostly provided by Commander. 

Should be enough

src/commands/service/create.test.ts
```diff
@@ -1,10 +1,11 @@
 import fs from "fs";
-import os from "os";
 import path from "path";
 import { promisify } from "util";
 import uuid from "uuid/v4";
 import { Bedrock } from "../../config";
+import * as config from "../../config";
 import { checkoutCommitPushCreatePRLink } from "../../lib/gitutils";
+import { createTempDir, removeDir } from "../../lib/ioUtil";
```
sarath-p, line 9: `createTempDir` is not implemented in this PR, but `createDir` is a preferable name and caller can decide how the dir being used.

src/commands/service/create.test.ts
```diff
@@ -25,38 +26,135 @@ afterAll(() => {
...
+  middlewares: "",
+  middlewaresArray: [],
+  packagesDir: "",
+  pathPrefix: "",
+  pathPrefixMajorVersion: "",
+  variableGroups: []
+};
+
+const getMockValues = (): ICommandValues => {
+  // TOFIX: if possible, can we use lodash?
```
sarath-p, line 56: You can create an issue to be prioritized and delete the comment.

sarath-p (approved):
Looks good.

src/commands/service/create.ts
evanlouie: for clarity, the reason we typecheck the the values from `opts` is because `opts` has a type of `{[key: string]: any}`. So when your create your `ICommandValues`, we are implicitly casting those values to whatever the key in `ICommandValues` points to.

src/commands/service/create.ts
```diff
@@ -10,328 +10,144 @@ import {
...
+    displayName: opts.displayName,
+    gitPush: opts.gitPush,
+    helmChartChart: opts.helmChartChart,
+    helmChartRepository: opts.helmChartRepository,
+    helmConfigBranch: opts.helmConfigBranch,
+    helmConfigGit: opts.helmConfigGit,
+    helmConfigPath: opts.helmConfigPath,
+    k8sBackend: opts.k8sBackend,
+    k8sBackendPort: opts.k8sBackendPort,
+    k8sPort: parseInt(opts.k8sBackendPort, 10),
```
evanlouie, line 75: re-add the logic from `isValidConfig` to check the value for `k8sBackendPort` is a positive integer (no floats, no negative numbers, and no `NaN`)

src/commands/service/create.ts
dennisseah: ok

src/commands/service/create.test.ts
dennisseah: `createTempDir` function already exist (added in previous PR)

src/commands/service/create.test.ts
dennisseah: ok

src/lib/validator.ts
```diff
@@ -15,6 +15,32 @@ export const hasValue = (val: undefined | null | string): val is string => {
   return val !== undefined && val !== null && val !== "";
 };
 
+/**
+ * Returns true of val is not undefined, not null and not an empty string
+ * and can be converted to integer.
+ *
+ * @param val Value to inspect
+ */
+export const isInteger = (val: undefined | null | string): val is string => {
```
evanlouie: when implementing type-guards, loosen the type constraint on `val` to be `unknown`:

```ts
 export const isInteger = (val: unknown): val is string => {
```

src/lib/validator.ts
```diff
@@ -15,6 +15,32 @@ export const hasValue = (val: undefined | null | string): val is string => {
...
+  }
+  return /^[1-9]\d+$/.test(val);
+};
+
+/**
+ * Returns true of val is a port number.
+ *
+ * @param val Value to inspect
+ */
+export const isPortNumber = (val: undefined | null | string): val is string => {
```
evanlouie: loosen the type constraint on `val` to be `unknown`

src/lib/validator.ts
evanlouie: also, this functions name is ambiguous. `isInteger` should correspond to the logic `Number.isInteger`. that is `type isInteger = (value: unknown) => value is number`.

The logic you've implemented should be named more aptly something like `isIntegerString()`

src/lib/validator.ts
evanlouie: Name should be more representative of functionality. Change to something like `isPortNumberString`

evanlouie (changes requested):

dennisseah:
@evanlouie, can you let me know what are the changes as you have a change requested. thanks

src/commands/service/create.ts
```diff
@@ -10,328 +10,144 @@ import {
...
+
+export interface ICommandValues extends ICommandOptions {
+  k8sPort: number;
+  middlewaresArray: string[];
+  variableGroups: string[];
+}
+
+export const fetchValues = (opts: ICommandOptions) => {
+  if (!isPortNumber(opts.k8sBackendPort)) {
+    throw new Error("value for --k8s-service-port is not a value port number");
```
bnookala: typo - maybe you meant `valid port number`?

src/commands/service/create.ts
dennisseah: at the beginning of the function, we went farther to make sure that the port range is valid.
```
if (!isPortNumber(opts.k8sBackendPort)) {
    throw new Error("value for --k8s-service-port is not a value port number");
}
```

src/commands/service/create.ts
```diff
@@ -10,328 +10,144 @@ import {
...
+  }
+
+  let variableGroups: string[] = [];
+
+  try {
+    const bedrock = Bedrock();
+    variableGroups = bedrock.variableGroups || [];
+  } catch (_) {
+    // NO-OP
```
bnookala: why is this a no-op?

src/lib/validator.ts
dennisseah: I have no problem changing the function name if you really think that it is necessary. Please let me know

src/lib/validator.ts
dennisseah: I have no problem changing the function name if you really think that it is necessary. Please let me know

src/lib/validator.ts
dennisseah: I do not see that reason for loosing the type because we precisely know the types from commander is doing to be either undefined, null or string.

src/lib/validator.ts
dennisseah: > loosen the type constraint on `val` to be `unknown`

or we can remove the type guard :-)

src/lib/validator.ts
evanlouie: - rename the function.
- we want to loosen the constraint on the function because you have written a general validation function that we should want to be able to reuse in other situations if need be (eg, if we want to use the function to parse values form config files or alike).

src/lib/validator.ts
evanlouie: - rename the function.
- no reason to not have a typeguard

src/commands/service/create.ts
dennisseah: sorry, I should have added some comments. For testing (running the command), I found that this command is ok without bedrock.yaml. the presence of this yaml file is to get the variable groups information, please let me know if this not accurate. That's bedrock.yaml file is needed and I will amend the changes. Otherwise I will add some comments like this 
```
bedrock.yaml file is optional. We get variable groups information in it if it exists
```

src/commands/service/create.ts
dennisseah: accept

src/lib/validator.ts

src/lib/validator.ts

src/commands/service/create.test.ts
dennisseah: https://github.com/microsoft/bedrock/issues/933

src/commands/service/create.ts
sarath-p: May be we need to extend test cases. We check if the user ran `project init` and `project cvg` commands before running this command so `bedrock.yaml` should exist with at least one variable group. The below line performs the validation.

https://github.com/CatalystCode/spk/blob/d1e0ac09ecb9e74808c02d57536facef640b20c3/src/commands/service/create.ts#L110

dennisseah: can you please create a task? And we can do this in a separate PR

sarath-p: That will be fine. I think we can remove try/catch block here since the code never reaches here without a `bedrock.yaml` file in the directory.

dennisseah: ok. I will remove the try-catch block. That's this command needs bedrock.yaml otherwise it will fail. thanks

evanlouie (approved):

PR #242, opened 2020-02-03
[REFACTOR] Refactor service pipeline ts

dennisseah:

2. standardized on the function in command ts file
3. use the CommandBuilder's exit function
4. default value of command option values is set to "" and values fall back to values in config file, typeof optionValue === "string" is not needed.
5. added more tests (hitting over 90% coverage)
6. avoid functions with too many parameters by using interface.

ran sh validations.sh successfully


andrebriggs (approved):
LGTM

bnookala (approved):

PR #245, opened 2020-02-04
Refactoring service build and update pipelines

### LINKED ISSUE #907: Issues in `spk install-build-pipeline` produced azure-pipelines.yaml file
**Describe the bug:**
I was able to get pass #906 issue by manually changing the AzDO pipeline path to the `azure-pipelines.yaml` file. Once that is done I can run a pipeline scaffolded by the `spk service install-build-pipeline` command. 

The problem is that the `azure-pipelines.yaml` file doesn't seem to handle non-mono repos. A concatenation error makes the Docker image name invalid.

Below you will see that `fabrikam-frontend-:master-20200126.1` is not a valid Docker name image. 

<pre>
/bin/bash --noprofile --norc /home/vsts/work/_temp/6d5772c7-942c-41c9-99de-ce19c9531f73.sh
<b>Image Name: fabrikam-frontend-</b>
az acr build -r andrebriggs --image <b>fabrikam-frontend-:master-20200126.1 .</b>
WARNING: Packing source code into tar to upload...
WARNING: Excluding '.gitignore' based on default ignore rules
WARNING: Excluding '.git' based on default ignore rules
WARNING: Uploading archived source code from '/tmp/build_archive_f15dc90c825941dfaf2161e7c06e8e28.tar.gz'...
WARNING: Sending context (2.010 KiB) to registry: andrebriggs...
<b>ERROR: Some of the properties of 'runRequest' are invalid.. InnerErrors: DockerBuildRequest:Image name 'fabrikam-frontend-:master-20200126.1' is invalid. 
The acceptable patterns are 'repository' or 'repository:tag'.The repository and tag names follow the standardized docker repository and tag naming conventions. Tag templates are supported within '{{' and '}}'. </b>
Please refer to docs for more information on using image name and tag:  https://aka.ms/az/acr/az-acr-build
##[error]Bash exited with code '1'.
Finishing: ACR Build and Publish
</pre>

The build worked once I removed the `-`. See below:

image

I assume that there was supposed to be something injected into the script after the `-`. Maybe that didn't happen in the non-mono repo situation

Another issue is the creation of a variable named `PROJECT_NAME_LOWER` in the `azure-pipelines.yaml` generated by the `spk service install-build-pipeline` command. Someone `PROJECT_NAME_LOWER`doesn't get populated and that results in a PR on a branch with two dashes. Example: 

image

```Merge DEPLOY/fabrikam-frontend--master-20200126.2 to refs/heads/master```

image

I fixed this by removing the bolded part below. This is certainly a hack

<pre>
 - script: |-
              export PROJECT_NAME_LOWER=$(echo  | tr '[:upper:]' '[:lower:]')
              export BUILD_REPO_NAME=$(echo $(Build.Repository.Name)<b>-$PROJECT_NAME_LOWER</b> | tr '[:upper:]' '[:lower:]')
              # --- From https://raw.githubusercontent.com/Microsoft/bedrock/master/gitops/azure-devops/release.sh
              . build.sh --source-only
</pre>

**To Reproduce:**

See #906 

**Expected behavior:**
Build from `azure-pipelines.yaml` should work and produce the correct PR on the HLD repo

**Additional context:**

### PR DESCRIPTION
**Pull request by mtarng**
Closes https://github.com/microsoft/bedrock/issues/907 and https://github.com/microsoft/bedrock/issues/801

Also moving pipeline definition file names into constants file.

Should work in conjunction with https://github.com/microsoft/bedrock/issues/904 to ensure image tags have proper names if `spk service create .` is called.

### DIFF
--- a/src/commands/project/pipeline.ts
+++ b/src/commands/project/pipeline.ts
@@ -12,10 +12,11 @@ import {
   validateForRequiredValues
 } from "../../lib/commandBuilder";
 import {
+  BUILD_SCRIPT_URL,
   PROJECT_CVG_DEPENDENCY_ERROR_MESSAGE,
-  PROJECT_INIT_CVG_DEPENDENCY_ERROR_MESSAGE
+  PROJECT_INIT_CVG_DEPENDENCY_ERROR_MESSAGE,
+  PROJECT_PIPELINE_FILENAME
 } from "../../lib/constants";
-import { BUILD_SCRIPT_URL } from "../../lib/constants";
 import {
   getOriginUrl,
   getRepositoryName,
@@ -148,14 +149,14 @@ const createPipeline = async (
   devopsClient: IBuildApi
 ): Promise<BuildDefinition> => {
   const definition = definitionForAzureRepoPipeline({
-    branchFilters: ["master"], // I believe this is intentional. Our pipelines are triggered only by merges into the master branch.
+    branchFilters: ["master"], // Pipelines are triggered only by merges into the master branch.
     maximumConcurrentBuilds: 1,
     pipelineName: values.pipelineName!,
     repositoryName: values.repoName!,
     repositoryUrl: values.repoUrl!,
     variables: requiredPipelineVariables(values.buildScriptUrl!),
-    yamlFileBranch: "master", // I believe this is intentional. Our pipelines are triggered only by merges into the master branch.
-    yamlFilePath: "hld-lifecycle.yaml" // TOFIX: hardcoded?
+    yamlFileBranch: "master", // Pipeline is defined in master
+    yamlFilePath: PROJECT_PIPELINE_FILENAME // Pipeline definition lives in root directory.
   });
 
   try {
--- a/src/commands/project/pipeline.ts
+++ b/src/commands/project/pipeline.ts
@@ -149,7 +149,7 @@ const createPipeline = async (
   devopsClient: IBuildApi
 ): Promise<BuildDefinition> => {
   const definition = definitionForAzureRepoPipeline({
-    branchFilters: ["master"], // Pipelines are triggered only by merges into the master branch.
+    branchFilters: ["master"], // hld reconcile pipeline is triggered only by merges into the master branch.
     maximumConcurrentBuilds: 1,
     pipelineName: values.pipelineName!,
     repositoryName: values.repoName!,
--- a/src/commands/service/create.ts
+++ b/src/commands/service/create.ts
@@ -171,7 +171,7 @@ export const createService = async (
   shelljs.mkdir("-p", newServiceDir);
 
   // Create azure pipelines yaml in directory
-  await generateServiceBuildAndUpdatePipelineYaml(
+  generateServiceBuildAndUpdatePipelineYaml(
     rootProjectPath,
     values.ringNames,
     serviceName,
--- a/src/commands/service/create.ts
+++ b/src/commands/service/create.ts
@@ -174,7 +174,7 @@ export const createService = async (
   generateServiceBuildAndUpdatePipelineYaml(
     rootProjectPath,
     values.ringNames,
-    values.displayName ?? serviceName, // displayName takes priority over serviceName (which could be '.')
+    values.displayName ? values.displayName : serviceName, // displayName takes priority over serviceName (which could be '.')
     newServiceDir,
     values.variableGroups
   );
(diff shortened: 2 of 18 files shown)

### CODE REVIEW

**Comment by andrebriggs** [comment]
@mtarng thanks for the pull request. Let's avoid pull requests that cover multiple issues. Moreover #801 seems like it was brought into the sprint out of band.

**Comment by mtarng** [comment]
So the refactor and bug fixes were pretty much all tied together. The issues stemmed from how we referenced our service names in our build pipelines. Originally we supported onboarding multiple services within a `packages` directory that were all built by a single pipeline. So removing that functionality and having our service names be passed by a specific argument was the fix as we were previously crawling directories and using each directory as the name.

@andrebriggs

File: src/lib/constants.ts [hunk 1]
```diff
@@ -1,6 +1,8 @@
 export const BUILD_SCRIPT_URL =
   "https://raw.githubusercontent.com/Microsoft/bedrock/master/gitops/azure-devops/build.sh";
 
+export const HLD_PIPELINE_FILENAME = "manifest-generation.yaml";
```
bnookala: can we call this RENDER_HLD_PIPELINE_FILENAME instead? Feeling like it could be confused with the PROJECT_PIPELINE_FILENAME below

src/commands/project/pipeline.ts
```diff
@@ -148,14 +149,14 @@ const createPipeline = async (
   devopsClient: IBuildApi
 ): Promise<BuildDefinition> => {
   const definition = definitionForAzureRepoPipeline({
-    branchFilters: ["master"], // I believe this is intentional. Our pipelines are triggered only by merges into the master branch.
+    branchFilters: ["master"], // Pipelines are triggered only by merges into the master branch.
```
bnookala: _only_ the reconcile pipeline/project pipeline, service build pipeline is triggered on ring branches

src/commands/service/create.ts
```diff
@@ -173,9 +171,13 @@ export const createService = async (
   shelljs.mkdir("-p", newServiceDir);
 
   // Create azure pipelines yaml in directory
-  await generateStarterAzurePipelinesYaml(rootProjectPath, newServiceDir, {
-    variableGroups: values.variableGroups
-  });
+  generateServiceBuildAndUpdatePipelineYaml(
+    rootProjectPath,
+    values.ringNames,
+    values.displayName ? values.displayName : serviceName, // displayName takes priority over serviceName (which could be '.')
```
bnookala: i know in reconcile we try to use displayName and ignore services that are named '.' - can we catch and do something about a service name being '.' ?

yradsmikham (approved):

dennisseah (approved):
I only can review the ts code and they LGTM

src/lib/fileutils.ts
```diff
@@ -2,100 +2,93 @@ import fs from "fs";
...
+  ringBranches: string[],
+  variableGroups?: string[]
+): IAzurePipelinesYaml => {
+  const relativeServicePathFormatted = relServicePath.startsWith("./")
+    ? relServicePath
+    : "./" + relServicePath;
+
   // tslint:disable: object-literal-sort-keys
-  const starter: IAzurePipelinesYaml = {
+  const piplineYaml: IAzurePipelinesYaml = {
```
bnookala: pipelineYaml

src/lib/fileutils.ts
```diff
@@ -304,26 +255,13 @@ export const starterAzurePipelines = async (opts: {
...
-    );
-  }
+  const spkServiceBuildPipelineCmd =
+    "spk service install-build-pipeline " + serviceName;
+  logger.info(
+    `Generated ${SERVICE_PIPELINE_FILENAME} for service in path '${relativeServicePathFormatted}'. Commit and push this file to master before attempting to deploy via the command '${spkServiceBuildPipelineCmd}'; before running the pipeline ensure the following environment variables are available to your project variable groups: ${requiredPipelineVariables}`
+  );
 
-  return starter;
+  return piplineYaml;
```
bnookala: pipelineYaml

bnookala (changes requested):
some nits, but otherwise lg2m

bnookala:
Also, not actually sure this closes https://github.com/microsoft/bedrock/issues/908 given the fab set invocation hasn't been updated: https://github.com/CatalystCode/spk/pull/245/files#diff-dff86319339dff9db12b61d65beced10R127

src/commands/service/create.ts
mtarng: That entry check will be done with this ticket: https://github.com/microsoft/bedrock/issues/904

@yradsmikham 


But I'll also add a boolean check and throw an error.

mtarng:
Removing microsoft/bedrock#908 , that was a different issue than I had thought.

bnookala (approved):

PR #248, opened 2020-02-04
Deprecate infra validate command

### LINKED ISSUE #937: SPK Infra validate unit test takes 8 seconds to run
**Describe the bug:**

image

Not sure if we are using the `spk infra validate` command at all. It is not exercised in the [integration tests](https://github.com/CatalystCode/spk/blob/master/tests/infra-validations.sh).

If not let's remove the test that accounts for 22% of the SPK unit test stage

yradsmikham:
Close https://github.com/microsoft/bedrock/issues/937.

Moving dependencies to `init.ts`, and then deprecating `spk infra validate`.


yradsmikham:
Tests are failing because there is a dependency on infra validate. So we can deprecate the infra validate tests, but not the actual command itself.

@andrebriggs

jmspring (commented):
Lgtm pending tests passing

src/commands/deployment/dashboard.ts
```diff
@@ -5,7 +5,7 @@ import { Config } from "../../config";
 import { getRepositoryName } from "../../lib/gitutils";
 import { exec } from "../../lib/shell";
 import { logger } from "../../logger";
-import { validatePrereqs } from "../infra/validate";
+import { validatePrereqs } from "./../init";
```
samiyaakhtar: maybe these should be `../../init`?

samiyaakhtar (approved):

src/commands/deployment/dashboard.ts
yradsmikham: I tested that the path `./../init` works, but also have condensed it `../init` now.

dennisseah (approved):
LGTM

jmspring (approved):
lgtm - tests passing

PR #249, opened 2020-02-04
spk init config file should handle multiple env vars

### LINKED ISSUE #947: spk init config file does not handle multiple env vars in the same value correctly
**Describe the bug:**
For the purposes of integration test, I want to be able to specify in the config file:
```
hld_repository: "https://dev.azure.com/${env:AZDO_ORG}/${env:AZDO_PROJECT}/_git/${env:hld_dir}"
```
When I run `spk init` with a config file that contains multiple values defined like above, it only replaces the first occurrence of an env variable. 

**To Reproduce:**

**Expected behavior:**
It should replace every occurrence with its corresponding env variable. 

**Screenshots:**

**Additional context:**

samiyaakhtar:
Closes https://github.com/microsoft/bedrock/issues/947

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -69,7 +69,7 @@ export const loadConfigurationFromLocalEnv = <T>(configObj: T): T => {
 export const updateVariableWithLocalEnv = (value: string): string => {
   const regexp = /\${env:([a-zA-Z_$][a-zA-Z_$0-9]+)}/g;
   const original = value;
-  let matches = regexp.exec(value as string);
+  let matches = regexp.exec(value);
   while (matches) {
     if (matches.length > 1) {
       if (process.env[matches[1]]) {
@@ -80,7 +80,7 @@ export const updateVariableWithLocalEnv = (value: string): string => {
           `Environment variable needs to be defined for ${matches[1]} since it's referenced in the config file.`
         );
       }
-      matches = regexp.exec(original as string);
+      matches = regexp.exec(original);
     }
   }
   return value;
--- a/src/config.ts
+++ b/src/config.ts
@@ -49,19 +49,7 @@ export const loadConfigurationFromLocalEnv = <T>(configObj: T): T => {
   const iterate = (obj: any) => {
     if (obj != null && obj !== undefined) {
       for (const [key, value] of Object.entries(obj)) {
-        const regexp = /\${env:([a-zA-Z_$][a-zA-Z_$0-9]+)}/g;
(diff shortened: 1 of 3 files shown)


samiyaakhtar:
Integration tests are failing since my forked repo can't access variables in the pipeline. Smoke tests for PR are a new thing? @andrebriggs

src/config.ts
```diff
@@ -73,6 +61,31 @@ export const loadConfigurationFromLocalEnv = <T>(configObj: T): T => {
...
+    if (matches.length > 1) {
+      if (process.env[matches[1]]) {
+        value = value.replace(matches[0], process.env[matches[1]] as string);
+      } else {
+        logger.error(`Env variable needs to be defined for ${matches[1]}`);
+        throw Error(
+          `Environment variable needs to be defined for ${matches[1]} since it's referenced in the config file.`
+        );
+      }
+      matches = regexp.exec(original as string);
```
dennisseah: [change request]
we may have a chance to get into an infinite loop. I thought this liner should be
```
matches = regexp.exec(value);
```

src/config.ts
```diff
@@ -73,6 +61,31 @@ export const loadConfigurationFromLocalEnv = <T>(configObj: T): T => {
...
 
+/**
+ * Updates env variable(s) from process.env
+ * Supports multiple per value
+ * @param value Value from config file to be replaced with env vars
+ */
+export const updateVariableWithLocalEnv = (value: string): string => {
+  const regexp = /\${env:([a-zA-Z_$][a-zA-Z_$0-9]+)}/g;
+  const original = value;
+  let matches = regexp.exec(value as string);
```
dennisseah: can be just 
```
let matches = regexp.exec(value);
```
because value is already a string

dennisseah (changes requested):

src/config.ts
samiyaakhtar: `value` is being updated inside the loop. Based on docs if you want to iterate through all matches with regex, you need to call the method on the original string over and over. It's not intuitive at all, agreed. ([reference](https://stackoverflow.com/a/432503))

dennisseah (approved):
LGTM

mtarng:
is there an associated unit test for this? @samiyaakhtar

samiyaakhtar:
@mtarng thanks for the reminder, added a quick one

mtarng (approved):

PR #251, opened 2020-02-04
[TEST] added basic tests for updateVariableWithLocalEnv

dennisseah:
was reviewing and testing updateVariableWithLocalEnv. Adding my test code as part of unit tests

### DIFF
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -2,9 +2,9 @@ import os from "os";
 import path from "path";
 import shell from "shelljs";
 import uuid from "uuid/v4";
-import { Bedrock, write } from "./config";
-import { disableVerboseLogging, enableVerboseLogging, logger } from "./logger";
-import { IBedrockFile, IBedrockFileInfo } from "./types";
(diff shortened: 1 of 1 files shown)


src/config.test.ts
```diff
@@ -16,6 +16,33 @@ afterAll(() => {
...
+        "${env:hello2} - ${env:hello} : ${env:hello1}"
+      );
+    } catch (e) {
+      expect(e.message).toBe(
+        "Environment variable needs to be defined for hello2 since it's referenced in the config file."
+      );
+    }
+  });
+});
+
```
samiyaakhtar, line 45: there's a similar unit test located in `init.test.ts` as well :)

samiyaakhtar (approved):

src/config.test.ts
dennisseah: I am hoping that the unit test is in config.test.ts (since the code is in config.ts) and we can do more test variations. :-) Thanks

andrebriggs (approved):

PR #265, opened 2020-02-06
Fix incorrect `fab set` call

### LINKED ISSUE #945: HLD update build step does not update the correct image tag
**Describe the bug:**

Committing a change to my application repository does _not_ update the proper component configuration in the HLD generated by reconcile

**To Reproduce:**

1. Set up any buildable application via spk, and create a HLD + HLD lifecycle pipeline
2. Commit to application repository on any ring branch
3. Observe that HLD update step does _not_ update HLD to reflect the container image built at the ring level in the HLD.

**Expected behavior:**
 
My HLD to have a Pull Request against it containing an image tag update against the proper subcomponent

**Additional context:**

https://github.com/CatalystCode/spk/blob/master/src/lib/fileutils.ts#L240 invokes `fab set` on $PROJECT_NAME_LOWER, which does not work as $PROJECT_NAME_LOWER does _not_ reflect the proper path of the chart within an HLD.

A correct invocation of `fab set` would be to update the helm component at the leaf node of the HLD ie:

`fab set --subcomponent "$REPOSITORY.$SERVICE_NAME.$RING.$CHART_NAME  image.tag=$(Build.SourceBranchName)-$(Build.BuildNumber)`

Yes

evanlouie:
Calls to `fab set` in generated pipeline YAML files have been using the
incorrect component path to set -- only using the `<service-name>` instead of the
full path in the HLD of
`<repository/application-name>.<service-name>.<ring-name>.chart`. This fix
corrects the outputted `fab set` call.

- Corrected the `fab set` call in service build/update pipeline to `set` the
  correct component path.
- Corrected some incorrect error messaging in gitutils.

fixes https://github.com/microsoft/bedrock/issues/945


src/lib/fileutils.ts
```diff
@@ -194,7 +194,7 @@ export const serviceBuildAndUpdatePipeline = (
                   ``,
                   `# Update HLD`,
                   `git checkout -b "$BRANCH_NAME"`,
-                  `../fab/fab set --subcomponent $SERVICE_NAME_LOWER image.tag=$(Build.SourceBranchName)-$(Build.BuildNumber)`,
+                  `../fab/fab set --subcomponent $(Build.Repository.Name).$SERVICE_NAME_LOWER.$(Build.SourceBranchName).chart image.tag=$(Build.SourceBranchName)-$(Build.BuildNumber)`,
```
bnookala: we might need to do some massaging of sourceBranchName - what if it was a git flow branch name (ie fix/some-issue-number or dev/foo-bar-baz)

bnookala (changes requested):

src/lib/fileutils.ts
bnookala: i think in that case we'd want something that transforms / to a - ie:

release/dev -> release-dev

src/test/mockFactory.ts
```diff
@@ -122,7 +122,7 @@ export const createTestServiceBuildAndUpdatePipelineYaml = (
                   ``,
                   `# Update HLD`,
                   `git checkout -b "$BRANCH_NAME"`,
-                  `../fab/fab set --subcomponent $SERVICE_NAME_LOWER image.tag=$(Build.SourceBranchName)-$(Build.BuildNumber)`,
+                  `../fab/fab set --subcomponent $(Build.Repository.Name).$SERVICE_NAME_LOWER.$(Build.SourceBranchName).chart image.tag=$(Build.SourceBranchName)-$(Build.BuildNumber)`,
```
andrebriggs: @evanlouie in the description you say 
> Calls to fab set in generated pipeline YAML files have been using the
incorrect component path to set -- only using the <service-name> instead of the
full path in the HLD of
<repository/application-name>-<service-name>-<ring-name>-chart. This fix
corrects the outputted fab set call.

However here we have `<application-name>-<service-name>-<ring-name> .chart` with a DOT chart. Is this correct?

andrebriggs (dismissed):

src/test/mockFactory.ts
evanlouie: whoops. Wrote the comment incorrectly. We delimit the parts with `.` not `-`.
So `<application-name>.<service-name>.<ring-name>.chart` is correct

mtarng:
Does this handle the case where service name contains `.`'s? ie. `fabrikam.acme.frontend`?

evanlouie:
> Does this handle the case where service name contains `.`'s? ie. `fabrikam.acme.frontend`?

@mtarng we do a one off escaping for the `fab set` call to do define the service name as `FAB_SAFE_SERVICE_NAME=$(SERVICE_NAME_LOWER | tr . - | tr / -)`. The rest of the usage of service name in the build&update pipeline use SERVICE_NAME_LOWER

mtarng (approved):
This melts my brain. typescript -> bash -> azdo pipeline variables.

src/lib/fileutils.ts
evanlouie: it now uses a SAFE_SOURCE_BRANCH var that is `$(echo $(Build.SourceBranchName) | tr / - | tr . -)` (all `/` and `.` are replaced with `-`)

bnookala (approved):

PR #270, opened 2020-02-07
[REFACTOR] Deployment validate cmd

dennisseah:

did not do much with the code, mainly adding tests

2. standardized on the function in command ts file
3. use the CommandBuilder's exit function
4. added tests (hitting over 90% coverage)
5. declare interface for input values
6. have validation functions to throw exception and not return true/false to be sure that exception are caught and command terminates


samiyaakhtar (approved):

andrebriggs (approved):
LGTM

PR #273, opened 2020-02-09
[FIX] Issue 789: spk doesn't emit non-zero exit code on error

### LINKED ISSUE #789: spk doesn't emit non-zero exit code on error
**Describe the bug:**

**To Reproduce:**
I run something that [causes a failure](https://github.com/microsoft/bedrock/issues/788):
`$ spk project create-variable-group fabrikam-vg -r andrebriggs -d https://dev.azure.com/abrig/spk/_git/fabrikam-hld -u $SP_APP_ID -p $SP_PASS -t $SP_TENANT --org-name $AZDO_ORG --project $AZDO_PROJECT --personal-access-token $ACCESS_TOKEN_SECRET`

<pre>
info:    Creating Variable Group from group definition 'abrig'
info:    Creating Variable Group abrig
info:    Attempting to create Variable Group in project 'spk'
(node:9130) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
info:    Successfully connected to Azure DevOps Task API!
error:   Failed to create Variable Group abrig
<b> Error: Variable group 'abrig' already exists.
error:   Error occurred while creating variable group
error:   Variable group 'abrig' already exists.</b>
$ <b><i>status=$?
$ echo $status
0</i></b>
</pre>

**Expected behavior:**
I expect to see a non-zero exit code when calling $? in bash
> ($?) Expands to the exit status of the most recently executed foreground pipeline.

See more [here](https://www.gnu.org/software/bash/manual/bash.html#Special-Parameters)

Related to https://github.com/microsoft/bedrock/issues/759  --> If we returned easily parsable return information in JSON we could have another way to tell the success of a command. Not necessarily the solution but would help in figuring this out during automation.

dennisseah:
closes https://github.com/microsoft/bedrock/issues/789

added a file listener to listen to activity on spk.log file. Terminate the the exit code once there are activity on this file.

Test 1:
1. remove spk.log
2. execute `project create-variable-group fabrikam-vg` this is an error case because of missing values
3. `echo $?` gives 1
4. verified that the spk.log content is correct

Test 2:
1. keep spk.log (do not remove it)
2. execute `project create-variable-group fabrikam-vg` this is an error case because of missing values
3. `echo $?` gives 1
4. verified that the spk.log content is correct

Test 2:
1. keep spk.log (do not remove it)
2. execute `project create-variable-group fabrikam-vg -h` this is a valid command
3. `echo $?` gives 0
4. verified that the spk.log content is correct


dennisseah:
@edaena , can you please help to test if this fix works on Ubuntu? thanks

andrebriggs (approved):
Looks good. @NathanielRose can you also see if this works on Ubuntu?

samiyaakhtar (approved):

NathanielRose (approved):
works on linux and windows

PR #276, opened 2020-02-09
[REFACTOR] spk variable-group create command

dennisseah:

small changes because major refactoring is not possible because unit test is missing
1. move command declaration out to a JSON file
2. calling `CommandBuilder.exit` function to terminate the command.

### DIFF
--- a/src/commands/variable-group/create.ts
+++ b/src/commands/variable-group/create.ts
@@ -1,8 +1,8 @@
-import { VariableGroup } from "azure-devops-node-api/interfaces/ReleaseInterfaces";
 import commander from "commander";
 import fs from "fs";
 import path from "path";
 import { readYaml } from "../../config";
+import { build as buildCmd, exit as exitCmd } from "../../lib/commandBuilder";
 import { IAzureDevOpsOpts } from "../../lib/git";
 import {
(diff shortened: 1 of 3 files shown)


andrebriggs:
LGTM. We definitely need to do https://github.com/microsoft/bedrock/issues/946

andrebriggs (approved):

src/commands/variable-group/create.ts
```diff
@@ -1,96 +1,78 @@
...
+      );
 
-        const accessOpts: IAzureDevOpsOpts = {
-          orgName,
-          personalAccessToken,
-          project
-        };
-        logger.debug(`access options: ${JSON.stringify(accessOpts)}`);
+      // type check
+      if (typeof orgName !== "undefined" && typeof orgName !== "string") {
```
sarath-p, line 36: Are these type checks still valid with command builder implementation?

sarath-p (approved):

src/commands/variable-group/create.ts
dennisseah: I will not make these changes because I think we are going to combine this one command with another one. :-)

PR #280, opened 2020-02-10
Exclude variables with default values in definitions

### LINKED ISSUE #811: `spk infra scaffold` should generate `definition.yaml` that use and exclude default values if provided in .tf files
**As a:**
A developer

**I want:**
`spk infra scaffold` to parse, **use**, and _exclude_ the variables with default values in the `definition.yaml` file, instead of inserting a dummy placeholder `<insert value>`.

**So that:**
`spk infra scaffold` can generate a more complete `definition.yaml`, if possible.

- [x] Parse and extract the default values of the `variables.tf`, if they exists.
- [x] Generate a `definition.yaml` that excludes variables that have default values. Variables without default values will appear in the `definition.yaml`.

**Acceptance Criteria:**
The `definition.yaml` should use default values. Variables with default values should not appear in the definition file. Variables that do not have default values in the Terraform template will have a placeholder of `<insert value>`.

Yes

yradsmikham:
Close https://github.com/microsoft/bedrock/issues/811.


dennisseah (approved):
LGTM

mtarng (approved):

src/commands/infra/scaffold.ts
```diff
@@ -301,6 +300,12 @@ export const generateClusterDefinition = (
     Object.keys(fields).forEach(key => {
       fieldDict[key] = fields[key] || "<insert value>";
```
NathanielRose: Question: Can we always expect `<insert value>`

mtarng: This logic will put `"<insert value>"` as the mapped value in the dict, if the current value is null or an empty string.

Though, if we're requring the same constant string here and a bit below, I'd prefer to see this value be a const. @yradsmikham

yradsmikham: @mtarng done!

PR #281, opened 2020-02-10
[Test] Add validation for infra spk.tfvars file from generated repo

### LINKED ISSUE #973: Improved validation of Generated repo to account for `spk.tfvars` file content
**As a:**
Developer
**I want:**
Improved reliable integration testing for SPK Infra Generation Pipeline
**So that:**
I can assure the my generated repo contains the correct spk.tfvr files in the submitted pull request

Currently, in our integration test, we simply check for the existence of a PR and the correct files in the Generated repo. We want improved coverage for validating this test by checking the contents of the `spk.tfvars` file
- [x] Improved validation of Generated repo to account for `spk.tfvars` file content

**Acceptance Criteria:**
Updated  `infra-validations.sh` 

**Additional context:**
Continuation of #872 & #919 

No

NathanielRose:
- Closes https://github.com/microsoft/bedrock/issues/973
- A quick test for integration test to validate spk.tfvars file to assure PR for generated repo correctly configured the file.
- Should we do a more adequate test? Output the terraform plan and run a verification. The ticket will need to be updated

### DIFF
--- a/tests/infra-validations.sh
+++ b/tests/infra-validations.sh
@@ -233,5 +233,8 @@ git checkout $pr_id
 file_we_expect=("variables.tf" "main.tf" "backend.tfvars" "spk.tfvars")
 validate_directory "$TEST_WORKSPACE/$infra_generated_dir/$infra_hld_project-generated/$infra_region" "${file_we_expect[@]}" >> $TEST_WORKSPACE/log.txt
 echo "PR for generated repo validated."
-
+# Validate the contents of the definition.yaml
+spkVars_test=$'rg_name = "test-rg"\nrg_location = "west us2"\n'
+validate_file "$TEST_WORKSPACE/$infra_generated_dir/$infra_hld_project-generated/$infra_region/spk.tfvars" $spkVars_test >> $TEST_WORKSPACE/log.txt
+echo "SPK.tfvars file in the generated repo validated."
 echo "Successfully reached the end of the infrastructure validations script."


yradsmikham (approved):

tests/infra-validations.sh
```diff
@@ -241,5 +241,8 @@ git checkout $pr_id
 file_we_expect=("variables.tf" "main.tf" "backend.tfvars" "spk.tfvars")
 validate_directory "$TEST_WORKSPACE/$infra_generated_dir/$infra_hld_project-generated/$infra_region" "${file_we_expect[@]}" >> $TEST_WORKSPACE/log.txt
 echo "PR for generated repo validated."
-
+# Validate the contents of the definition.yaml
+spkVars_test=$'rg_name = "test-rg"\nrg_location = "west us2"\n'
+validate_file "$TEST_WORKSPACE/$infra_generated_dir/$infra_hld_project-generated/$infra_region/spk.tfvars" $spkVars_test >> $TEST_WORKSPACE/log.txt
```
andrebriggs, line 246: @NathanielRose I looked at `validate_file` and it is just a grep command with the quiet flag `grep -q "$2" "$1"` If there is more an $2 in the file it looks like we give success?

NathanielRose: @andrebriggs The reason why we use the grep command to check if a validation string is contained inside the file is due to line [127](https://github.com/CatalystCode/spk/blob/c0fcfec01f169d059cafe72f690855e14fd95484/tests/infra-validations.sh#L127). 

The definition yaml that gets created that we validate contains the PAT token and is a much larger file then the terraform file we validate at the end of the file. In the script, we check to see if a [validation string](https://github.com/CatalystCode/spk/blob/c0fcfec01f169d059cafe72f690855e14fd95484/tests/infra-validations.sh#L40) is within the definition.yaml. 

I can move forward to modify the function for `validate_file` to check the entire contents of the file and host it as a string in memory or a mock file. Or can I keep it as is with checking if a subset string exists in the file?

dennisseah: IMO, this validation is weak and it is ok because the `definition.yaml` file is created by `spk infra scaffold` and there are already unit test for `generateClusterDefinition` function

dennisseah (approved):

tests/infra-validations.sh
NathanielRose: Agreed with @dennisseah but since we have a validation that checks the files of the PR, and our unit tests, I think this will suffice for checking the contents of the `spk.tfvars` file. We can revisit in the future.

PR #284, opened 2020-02-11
Making sure middleware and ingressrouth paths are the same

### LINKED ISSUE #959: HLD Reconcile: generated middleware strips the wrong path
**Describe the bug:**

A generated middleware yaml from `hld reconcile` does not strip the full path that maps to the Ingress Route

**To Reproduce:**

Using this bedrock.yaml:

```
rings:
  master:
    isDefault: true
services:
  ./services/fabrikam:
    displayName: 'fabrikam'
    k8sBackendPort: 80
    k8sBackend: 'fabrikam-k8s-svc'
    pathPrefix: 'fabrikam-service'
    pathPrefixMajorVersion: 'v1'
    helm:
      chart:
        branch: master
        git: 'https://dev.azure.com/foo/bar/_git'
        path: stable/wordpress
    middlewares:
      - ''
variableGroups:
  - fabrikam-vg
```

Running `spk hld reconcile` on an application repository produces this middleware (note the prefixes):

```
kind: Middleware
metadata:
  name: fabrikam-master
spec:
  stripPrefix:
    forceSlash: false
    prefixes:
      - /fabrikam
```

and this ingress route (note the path path match):

```
kind: IngressRoute
metadata:
  name: fabrikam-master
spec:
  routes:
    - kind: Rule
      match: 'PathPrefix(`/v1/fabrikam-service`) && Headers(`Ring`, `master`)'
      middlewares:
        - name: fabrikam-master
      services:
        - name: fabrikam-k8s-svc
          port: 80
```

**Expected behavior:**

The generated middleware to utilize the path prefix and the pathPrefixMajorVersion as a prefix strip ie:

```
kind: Middleware
metadata:
  name: fabrikam-master
spec:
  stripPrefix:
    forceSlash: false
    prefixes:
      - /v1/fabrikam-service
```

**Additional context:**

It looks like the middleware presently uses just the display name of the service

No

mtarng:
Closes https://github.com/microsoft/bedrock/issues/959

### DIFF
--- a/src/commands/hld/reconcile.ts
+++ b/src/commands/hld/reconcile.ts
@@ -5,6 +5,7 @@ import yaml from "js-yaml";
 import path from "path";
 import process from "process";
 import shelljs, { TestOptions } from "shelljs";
+import url from "url";
 import { Bedrock } from "../../config";
 import { assertIsStringWithContent } from "../../lib/assertions";
 import { build as buildCmd, exit as exitCmd } from "../../lib/commandBuilder";
@@ -350,10 +351,10 @@ export const getFullPathPrefix = (
   pathPrefix: string,
   serviceName: string
 ): string => {
-  const versionPath = majorVersion ? `/${majorVersion}` : "/";
+  const versionPath = majorVersion ? `/${majorVersion}` : "";
   const servicePath = pathPrefix || serviceName;
 
-  return path.join(versionPath, servicePath);
+  return `${versionPath}/${servicePath}`;
 };
 
 /**
--- a/src/commands/hld/reconcile.ts
+++ b/src/commands/hld/reconcile.ts
@@ -116,13 +116,15 @@ export interface IReconcileDependencies {
     serviceName: string,
     serviceConfig: IBedrockServiceConfig,
     middlewares: ITraefikMiddleware,
-    ring: string
+    ring: string,
+    ingressVersionAndPath: string
   ) => void;
 
   createMiddlewareForRing: (
     ringPathInHld: string,
     serviceName: string,
-    ring: string
+    ring: string,
+    ingressVersionAndPath: string
   ) => ITraefikMiddleware;
 }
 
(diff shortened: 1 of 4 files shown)


src/commands/hld/reconcile.ts
```diff
@@ -309,25 +311,51 @@ export const reconcileHld = async (
...
+ */
+export const getFullPathPrefix = (
+  majorVersion: string,
+  pathPrefix: string,
+  serviceName: string
+): string => {
+  const versionPath = majorVersion ? `/${majorVersion}` : "/";
+  const servicePath = pathPrefix || serviceName;
+
+  return path.join(versionPath, servicePath);
```
dennisseah: I thought this should be
```
  const versionPath = majorVersion ? `/${majorVersion}` : "";
  const servicePath = pathPrefix || serviceName;
  return `${versionPath}/${servicePath}`;
```
this is URL path and not file system path (if i am not wrong)

bnookala: @dennisseah I think @mtarng's fix should be fine - we're specifying the full path to strip (from the root) of the url ie:

http://bar.com/v1/fabrikam should strip /v1/fabrikam

bnookala: and even then:

```
> path.join("/", "foo")
'/foo'
```

bnookala (approved):

src/commands/hld/reconcile.ts
dennisseah: `path` library should not be used on constructing URL. separator is always `/` is there a reason for using `path`? Thanks

yradsmikham (approved):

src/commands/hld/reconcile.ts
dennisseah: https://stackoverflow.com/questions/16301503/can-i-use-requirepath-join-to-safely-concatenate-urls

edaena (approved):

PR #290, opened 2020-02-12
Introspection onboard command fails on config update

### LINKED ISSUE #982: spk introspection smoke test is failing due to azure table deletion step
**Describe the bug:**
Introspection smoke test is failing because the table we create during the test is not being successfully deleted by azure. We don't need to "test azure" in spk smoke test. 

**To Reproduce:**
https://dev.azure.com/epicstuff/bedrock/_build/results?buildId=11737&view=logs&j=75b9d1eb-f58e-5019-295a-7684b5b743da&t=06321f81-91d7-59ce-8241-0a1a98595180

**Expected behavior:**
Test should only be focussing on spk 

**Screenshots:**
Screen Shot 2020-02-11 at 4 44 10 PM

**Additional context:**

samiyaakhtar:
Screen Shot 2020-02-12 at 12 52 44 PM

The goal of onboard command is to create a storage, not necessarily fail if writing to config file fails since the first time user will not have created a config file in 90% of the cases. A logger message is sufficient. 

Also, removed logging of access key

Closes https://github.com/microsoft/bedrock/issues/982


dennisseah (approved):

src/commands/deployment/onboard.ts
```diff
@@ -319,6 +316,6 @@ export const setConfiguration = (
     logger.error(
       `Unable to set storage account and table names in configuration file. \n ${err}`
     );
-    throw err;
+    return false;
```
andrebriggs, line 319: @samiyaakhtar how will the integration test interact with this now. If previously we threw an error will we check for a boolean result?

andrebriggs (changes requested):

src/commands/deployment/onboard.ts
samiyaakhtar: @andrebriggs The integration test is not trying to test whether config was written back to file or not (which is what this function is doing), since the onboard command is meant to be ran before a user has setup their config file. Failing the execution just because they don't have an existing config file isn't a good experience. So the integration test is not going to verify if writing to config succeeded because there is no init file setup in the build pipeline

samiyaakhtar: @andrebriggs it will however leave a logger message but not fail the test

NathanielRose (approved):
Lgtm, works on linux/win

src/commands/deployment/onboard.ts
andrebriggs: OK, I suspect we will rework the onboard command very soon

andrebriggs (approved):

PR #295, opened 2020-02-12
Spk deployment validate fix for verifying storage account credentials

### LINKED ISSUE #952: spk deployment validate should not fail on missing optional params in config
**Describe the bug:**
`spk deployment validate` should not perform a check against optional variables in the config. For eg. when I don't specify tenant id, SP Id, SP secret and subscription Id, it shouldn't fail. The configuration for accessing storage should only matter.

Only the following config args are required for introspection:
- `account_name`
- `table_name`
- `partition_key`
- `key`
- `manifest_repository` under `azure`

**To Reproduce:**
Run `spk deployment validate`

**Expected behavior:**
It should not fail if storage config is valid.

**Screenshots:**
Screen Shot 2020-02-04 at 2 20 29 PM

**Additional context:**

NathanielRose:
- Closes https://github.com/microsoft/bedrock/issues/952
- No longer need to verify storage account using tenant ID and Sub ID
- To test run `spk deployment validate`

### DIFF
--- a/src/commands/deployment/validate.test.ts
+++ b/src/commands/deployment/validate.test.ts
@@ -8,7 +8,6 @@ import {
   deleteSelfTestData,
   execute,
   isValidConfig,
-  isValidStorageAccount,
   runSelfTest,
   writeSelfTestData
 } from "./validate";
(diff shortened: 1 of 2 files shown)


samiyaakhtar (approved):
LGTM. Looks like code coverage variance test is unhappy

andrebriggs (approved):

PR #296, opened 2020-02-13
Fix broken hld_update condition

### LINKED ISSUE #988: hld_update stage not being triggered for non-master branch
**Describe the bug:** 

The `hld_update` stage of the service build pipeline is conditioned on:

```yaml
condition: and(succeeded('build'), or(startsWith(variables['Build.SourceBranch'], 'refs/heads/DEPLOY/'),eq(variables['Build.SourceBranchName'],'master')))
```

This constraint for the stage is is not required as the pipeline itself already utilizes branch triggers.

This breaks my pipeline as If I have defined it to trigger on branch/ring `prod`, the `hld_update` stage will never trigger as it doesn't meet the condition.

**To Reproduce:**
- Setup an spk project
  - Add a `prod` ring entry in `bedrock.yaml`
- Add a service
- Install the pipelines
- Notice that pushes to `prod` branch will not trigger the `hld_update` stage of the build pipeline

**Expected behavior:**

The pushed ACR image's tag is pushed to my HLD repository

**Screenshots:**

**Additional context:**

no

evanlouie:
The `hld_update` stage of the service build pipeline was conditional on the
pipeline source branch either being prefixed with `deploy` or being `master`.
This condition is not needed as the pipeline itself already contains branch
triggers.

closes https://github.com/microsoft/bedrock/issues/988


andrebriggs:
@evanlouie how would you verify this behavior in a test?

samiyaakhtar (approved):

evanlouie:
> @evanlouie how would you verify this behavior in a test?

@andrebriggs You need to test in an actual pipeline. The test would be to have a ring not named `master` or prefixed with `deploy/` and push to it. Right now the the first stage of the ACR build step runs (build/push to ACR), but the second stage will be skipped. With this, as long as the first stage is triggered and succeeds, the second stage will should always trigger.

andrebriggs:
@evanlouie can you write a task to include this scenario in our integration tests. Otherwise looks good!

yradsmikham (approved):

bnookala (approved):

andrebriggs:
@bnookala did you create the task to track this since the PR was merged? 

> @evanlouie can you write a task to include this scenario in our integration tests. Otherwise looks good!

PR #298, opened 2020-02-13
Fix issue 990: -v alias should be used in command line because it is ...

### LINKED ISSUE #990: -v alias should be used in command line because it is reserved by commander
**Describe the bug:**
we have an option `-v, --version <repository version>` in `spk scaffold`. Using `-v` here is not allowed because `-v` is used globally by commander for setting verbose mode on.

**To Reproduce:**
run `spk infra scaffold -s http://some.domain.com -v 0.1 -n test -t <template value>` will result in error.
`spk infra scaffold -s http://some.domain.com --version 0.1 -n test -t <template value>` is ok. Therefore using `-v` does not work

**Expected behavior:**

**Screenshots:**

**Additional context:**

dennisseah:
…reserved by commander

change `-v` alias to `-r`.
Do not need to update documentation because in all documents, we mentioned `--version`

closes https://github.com/microsoft/bedrock/issues/990


yradsmikham (approved):

yradsmikham:
surprised we haven't caught this bug earlier on. Thanks @dennisseah !

NathanielRose (approved):
Good Catch

PR #308, opened 2020-02-14
Check if pipelines yaml exists in repo

edaena:
[ Check whether `azure-pipeline.yaml` file exist in the repo or not before creating the build definition #951](https://github.com/microsoft/bedrock/issues/951)


src/commands/hld/pipeline.ts
```diff
@@ -82,6 +82,15 @@ export const execute = async (
 ) => {
   try {
     populateValues(opts);
+    const hasPipelineFile = await repositoryHasFile("manifest-generation.yaml");
```
mtarng: Can you use the constants for the filename?

https://github.com/CatalystCode/spk/blob/master/src/lib/constants.ts#L17-L21

src/commands/hld/pipeline.ts
```diff
@@ -82,6 +82,15 @@ export const execute = async (
...
+
+    if (!hasPipelineFile) {
+      logger.error(
+        "Error installing manifest generation pipeline. Repository does not have a manifest-generation.yaml file."
```
mtarng: Ditto.

src/commands/service/pipeline.ts
```diff
@@ -61,6 +62,17 @@ export const execute = async (
 ) => {
   try {
     await fetchValues(serviceName, opts);
+    const pipelineFileName = serviceName + "/azure-pipelines.yaml";
```
mtarng: Ditto. Also service pipeline is now named `build-update-hld.yaml`

mtarng (changes requested):

mtarng:
Also, i'm not sure this logic would work for all cases. The `install-*-pipeline` commands will point to whatever azdo org/project/repository that's configured for spk during `spk init` OR it can be pointed to whatever azdo org/project/repository via flags. There's no guarantee these commands need to be run in the root of a bedrock project repo. This is also true for the hld to manifest pipeline.

That being said, perhaps we could try a curl/get/fetch against the pipeline files, since we can construct the file path from the azdo org/project/repository/path/branch/filenames...?

mtarng:
@edaena https://docs.microsoft.com/en-us/rest/api/azure/devops/git/items/get?view=azure-devops-rest-5.1 

`GET https://dev.azure.com/{organization}/{project}/_apis/git/repositories/{repositoryId}/items?path={path}&api-version=5.1` maybe?

Be sure to include a branch. Should be master by default, but I'm currently working on an issue to install pipelines on other branches. https://github.com/microsoft/bedrock/issues/989

src/commands/hld/pipeline.ts
```diff
@@ -83,6 +85,29 @@ export const execute = async (
...
+    );
+
+    if (!hasPipelineFile) {
+      logger.error(
+        "Error installing build pipeline. Repository does not have a " +
+          RENDER_HLD_PIPELINE_FILENAME +
+          " file."
+      );
+      await exitFn(1);
+    }
```
dennisseah: you need a else here that's
```
if (!hasPipelineFile) {
...
} else {
   await installHldToManifestPipeline(opts);
   await exitFn(0);
}
```

otherwise exitFn will be called twice.

src/commands/project/pipeline.ts
```diff
@@ -126,6 +128,26 @@ export const execute = async (
...
+      );
+
+      if (!hasPipelineFile) {
+        logger.error(
+          "Error installing build pipeline. Repository does not have a " +
+            PROJECT_PIPELINE_FILENAME +
+            " file."
+        );
+        await exitFn(1);
+      }
```
dennisseah: same here. you need an else here

src/commands/service/pipeline.ts
```diff
@@ -62,6 +65,29 @@ export const execute = async (
...
+
+    if (!hasPipelineFile) {
+      logger.error(
+        "Error installing build pipeline. Repository does not have a " +
+          SERVICE_PIPELINE_FILENAME +
+          " file."
+      );
+      await exitFn(1);
+    }
+
```
dennisseah: same here, `else` is needed. 
Can you please create a helper function for this hasPipelineFile check because the same code is copied 3 times?

edaena: Sure, I can put all the logic in azdoClient instead of only the piece that checks if the file is present.

dennisseah: maybe `azdoutil.ts` is ok too

sarath-p: I would suggest using `azdoClient.ts` since we are planning to remove [`azdoutil.ts`](https://github.com/CatalystCode/spk/blob/4e48b6a28f9b64d196b41ea7358c5e17d4c2b832/src/lib/azdoutil.ts#L6) which has one duplicate function that also exist in [`azdoClient.ts`](https://github.com/CatalystCode/spk/blob/4e48b6a28f9b64d196b41ea7358c5e17d4c2b832/src/lib/azdoClient.ts#L19).

edaena: Yeah, there's very little in azdoutil, and I already have shared functionality in azdoClient as part of this pr

dennisseah (approved):

mtarng (approved):

sarath-p (approved):

PR #312, opened 2020-02-15
[FIX] issue 999: error in generate.ts is masked and hence command lin…

### LINKED ISSUE #999: error in generate.ts is masked and hence command line is returning status 0
**Describe the bug:**
`generateConfig` function is returning err object and not throwing it. hence `execute` function does not have the chance to catch the exception and return 0.

**To Reproduce:**
Bug is found via code inspection and this problem can be simulated by unit test.
Unit test will be part of the PR to fix this problem to prevent regression.

**Expected behavior:**
command tool to return status code 1

**Screenshots:**

**Additional context:**

No

dennisseah:
…e is returning status 0

Closes https://github.com/microsoft/bedrock/issues/999

modified code to not catch the exception and return it. Instead allow the exception to propagate to the calling function

### DIFF
--- a/src/commands/infra/generate.test.ts
+++ b/src/commands/infra/generate.test.ts
@@ -66,6 +66,31 @@ describe("fetch execute function", () => {
     expect(exitFn).toBeCalledTimes(1);
     expect(exitFn.mock.calls).toEqual([[1]]);
   });
+  it("negative time, simulate error in generateConfig function", async () => {
+    jest
+      .spyOn(generate, "validateDefinition")
+      .mockReturnValueOnce(DefinitionYAMLExistence.PARENT_ONLY);
+    jest.spyOn(generate, "validateTemplateSources").mockReturnValueOnce({});
+    jest
+      .spyOn(generate, "validateRemoteSource")
+      .mockReturnValueOnce(Promise.resolve());
+    jest
+      .spyOn(infraCommon, "getSourceFolderNameFromURL")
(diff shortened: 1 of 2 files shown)


andrebriggs (approved):

andrebriggs:
@dennisseah Tests are failing

dennisseah:
> @dennisseah Tests are failing

Yes, I need help from Nate.

dennisseah:
> > @dennisseah Tests are failing
> 
> Yes, I need help from Nate.

we have figured out the problem and the problem is in the unit test and our implementation is masking the issue. Hence fixing the implementation causes problem because unit test is incorrect. I have moved the git clone tests to the beginning of the test run and hence `~/.spk/templates/_yradsmikham_spk_source/` is available in the tests later. tests in PR passed now.

src/commands/infra/generate.test.ts
```diff
@@ -52,6 +52,147 @@ beforeEach(() => {
   jest.restoreAllMocks();
 });
 
+//////////////////////////////////////////////////////////////////////////////
```
NathanielRose, line 55: Did you mean to keep this?

NathanielRose (approved):
Integration tests and unit tests passing locally.

src/commands/infra/generate.test.ts
dennisseah: to clearly separate git related tests from other tests

PR #316, opened 2020-02-19
Fix issue 1004: some implementation issues with infra generate command

### LINKED ISSUE #1004: some implementation issues with infra generate command
**As a:**
developer

**I want:**
to have no duplicated code, no dead code

**So that:**
code quality is better

1. https://github.com/CatalystCode/spk/blob/master/src/commands/infra/generate.ts#L352
   We are already checking the both parent and leaf configuration exist.
   In lines https://github.com/CatalystCode/spk/blob/master/src/commands/infra/generate.ts#L390
   we check if parent configuration exist. This condition is always true. Hence the else block
   from line https://github.com/CatalystCode/spk/blob/master/src/commands/infra/generate.ts#L413
   to https://github.com/CatalystCode/spk/blob/master/src/commands/infra/generate.ts#L435
   are dead code. and having dead code causes maintenance issue.
2. line https://github.com/CatalystCode/spk/blob/master/src/commands/infra/generate.ts#L366
   to https://github.com/CatalystCode/spk/blob/master/src/commands/infra/generate.ts#L383
   is copy to https://github.com/CatalystCode/spk/blob/master/src/commands/infra/generate.ts#L441
   to https://github.com/CatalystCode/spk/blob/master/src/commands/infra/generate.ts#L457
   we have over 20 lines of duplicated code.
3. and the generateConfig function is over 100 lines long hence it is difficult to read and test
4. https://github.com/CatalystCode/spk/blob/master/src/commands/infra/generate.ts#L351
```
const childDirectory = path.join(parentDirectory, projectPath);
``` 
should be
```
const childDirectory = (parentPath === projectPath) ?
  parentDirectory :
  path.join(parentDirectory, projectPath);
``` 
and the unit test for (parentPath === projectPath) is missing.

**Acceptance Criteria:**
the above 4 issues are addressed and missing unit tests are added.

**Additional context:**

No

dennisseah:
closes https://github.com/microsoft/bedrock/issues/1004

1. break `generateConfig` function into small parts so that they can be easily tested
2. address issues mentioned in https://github.com/microsoft/bedrock/issues/1004
3. boost code coverage to 95%.


NathanielRose (approved):
lgtm, Thanks! for this @dennisseah

andrebriggs (approved):

PR #326, opened 2020-02-21
Fix broken service build trigger

### LINKED ISSUE #1016: build pipelines for services located in root do not trigger on changes in root directory
**To Reproduce:**
- Create a service in the root directory. ie. `spk service create . -n my-service`.
- Commit and deploy its build pipeline.
- After first build is done, edit something in the root directory like the Dockerfile.
- Check if the pipeline triggers.

**Expected behavior:**
Changing code in the root directory of the project should cause the service build pipeline to trigger.

To fix: for the build pipeline definition for services created in the root directory, do not include
```
  paths:
    include:
      - ./
```
under the trigger.

**Screenshots:**

**Additional context:**
Path has to match this regex: `^[^\/~\^\: \[\]\\]+(\/[^\/~\^\: \[\]\\]+)*$`

No

evanlouie:
- Service pipeline build triggers were set to invalid value of `./` when the
  service was in project root directory -- In such a case, we now longer include
  the path trigger.

fixes https://github.com/microsoft/bedrock/issues/1016


mtarng (approved):

src/lib/fileutils.test.ts
```diff
@@ -208,6 +208,35 @@ describe("generateServiceBuildAndUpdatePipelineYaml", () => {
     );
     expect(writeSpy).toBeCalled();
   });
+
+  test("no path trigger injected when the path is the project root (is: ./)", async () => {
```
dennisseah: can you remove `async`?

dennisseah (approved):

bnookala (approved):

PR #328, opened 2020-02-22
[HOUSEKEEPER] moving docs to md files

dennisseah:
Related to https://github.com/microsoft/bedrock/issues/976


src/commands/service/create.md
```diff
@@ -0,0 +1,25 @@
+## NOTE
```
mtarng: Seems like we should at least keep the Description portion of this command, even if it's a repeat from what's output by `spk service create -h`, just so things are consistent.

dennisseah: ok

andrebriggs (approved):

mtarng (approved):

PR #330, opened 2020-02-24
[HOUSEKEEPING] remove async in functions in tests

### LINKED ISSUE #1019: unit tests are taking longer to run because of async functions
**As a:**
Developer

**I want:**
to have unit tests run as fast as possible

**So that:**
I can save time

remove unnecessary async functions. Many functions can synchronous yet we have `async` for these functions

**Acceptance Criteria:**

**Additional context:**

No

dennisseah:
majority of the changes is to remove async in functions in tests. They are not needed and they slow down the tests execution. On my laptop, the tests run in 22 seconds; and after the change they ran in less than 10 seconds. These number may not be significant. However time is wasted for no reasons. As we add more tests, more delays may be introduced if this is not dealt with.

Closes https://github.com/microsoft/bedrock/issues/1019


andrebriggs (approved):
LGTM! Thanks for identifying some time savings

mtarng (approved):

PR #331, opened 2020-02-24
Fixed Broken getRemoteUrl Test

evanlouie:
Fixed a broken test that didn't work if you cloned the repository via SSH (as it
expected https). Have loosened it to do a pattern match instead.


mtarng (approved):

dennisseah (approved):
thanks

bnookala (approved):

PR #333, opened 2020-02-24
Increase timeout for service introspection onboard tests

edaena:


dennisseah (approved):
look good to me, I cannot figure out a better way to handle this than increasing the wait time. We may need to monitor if they work.

samiyaakhtar (approved):

PR #344, opened 2020-02-26
standardize on --devops-project with alias -d

dennisseah:
Related to  https://github.com/microsoft/bedrock/issues/1050

### DIFF
--- a/src/commands/project/create-variable-group.decorator.json
+++ b/src/commands/project/create-variable-group.decorator.json
@@ -29,7 +29,7 @@
       "required": true
     },
     {
-      "arg": "--org-name <organization-name>",
+      "arg": "-o, --org-name <organization-name>",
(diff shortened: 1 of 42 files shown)


sarath-p (approved):
Looks good.

src/commands/project/create-variable-group.decorator.json
```diff
@@ -9,7 +9,7 @@
       "required": true
     },
     {
-      "arg": "-d, --hld-repo-url <hld-repo-url>",
+      "arg": "-U, --hld-repo-url <hld-repo-url>",
```
edaena, line 12: why is this uppercase U?

dennisseah: cannot use -d because -d is for --devops-project. So I have to choose a new one. The plan to use -U for all url related options. And cannot use -u because -u is for user name or user id, sp id related command.

andrebriggs (approved):

PR #349, opened 2020-02-28
Spk setup

dennisseah:
this is the first implementation of spk setup command
related to https://github.com/microsoft/bedrock/issues/986

Note:
1. its command decoration json file has a disabled=true property so that the doc generator will skip it.
2. to test this command out, you need to uncomment a liner in index.ts
3. there are common code between spk init and this command; and these common code/functions are moved to lib folder

Next step: create HLD and Manifest Repo


andrebriggs (approved):
LGTM

src/commands/setup.ts
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * @param coreAPI Core API service
+ * @param name Name of Project
+ */
+export const createProject = async (coreAPI: ICoreApi, name: string) => {
+  try {
+    await coreAPI.queueCreateProject({
+      capabilities: {
+        processTemplate: {
+          templateTypeId: "27450541-8e31-4150-9947-dc59f998fc01" // TOFIX: do not know what this GUID is about
```
samiyaakhtar, line 138: Why is this guid needed?

src/index.ts
```diff
@@ -45,6 +46,7 @@ const commandModules = [
         c.version(require("../package.json").version);
       },
       initCommandDecorator
+      // setupCommandDecorator
```
samiyaakhtar, line 49: is this meant to be commented out?

src/commands/setup.ts
dennisseah: this is something that I cannot figure out its value; and I googled for it and found that this value is used. I asked Andre to found a contact where we can ask.

src/index.ts
dennisseah: yes, this command is not ready hence we are not registering it

samiyaakhtar: Got it, thank you

src/commands/setup.ts
samiyaakhtar: Got it

samiyaakhtar (approved):

PR #350, opened 2020-02-28
[Doc] added code to support showing command docs for each patch & master

### LINKED ISSUE #1063: Documentation site should handle multiple versions
**Describe the bug:**

If I use the docs site as a reference, it only contains documentation for the `master` branch of spk, instead of documentation for commands by releases

**To Reproduce:**

Browse to https://catalystcode.github.io/spk/commands/#init, and observe that commands and flags do not always line up with the latest release of `spk`.

**Expected behavior:**

I should be able to select a version of `spk` for which to view docs.

**Additional context:**

- Github pages always renders the documents in the `master` branch in the `docs` folder. If I am using a version of `spk` that is not built from the latest commit to `master` (ie any build that can be downloaded from the releases page), it is highly likely that I will be referencing docs that are irrelevant to my build of `spk`. 

No

Related to #976

dennisseah:
1. added a select control to select the version (release) of the current command line doc. default is master
2. added code to has version (release) in hash tag of URL so that it can be permanent link targeting different versions of doc
3. added code in `release-version-bump.sh` to create a new `data.json` for each patch, release; and update a file `releases.txt` so that UI is aware of the list of support releases.

closes https://github.com/microsoft/bedrock/issues/1063


dennisseah:
Screen Shot 2020-02-28 at 9 14 58 AM

andrebriggs (approved):
Looks great!

samiyaakhtar (approved):

PR #351, opened 2020-02-28
[HOUSEKEEPING] Standardize aliases for url options

dennisseah:
Related to https://github.com/microsoft/bedrock/issues/1050

1. use -U for url related option
2. use -M for --manifest-url because -U is already used in the same command
3. use -B for --build-script-url


andrebriggs (approved):
LGTM

docs/commands/data.json
```diff
@@ -271,12 +271,12 @@
         "defaultValue": ""
       },
       {
-        "arg": "-u, --hld-url <hld-url>",
+        "arg": "-U, --hld-url <hld-url>",
```
evanlouie, line 274: Its pretty rare and confusing when cli's use case sensitive flags. I think we should continue using lowercase.

evanlouie: thoughts @andrebriggs @bnookala ?

evanlouie (changes requested):

docs/commands/data.json
bnookala: yeah, not a fan of using uppercase flag names here… - is there any reason for them?

samiyaakhtar: Isn't the standard for output usually lowercase ` -o --output`? it's causing me some confusion on the get command now since only output is uppercased: 

```
Usage: deployment get|g [options]

Get the list of deployments and filter with these options: service name, environment, build ID, commit ID, container image tag.

Options:
  -b, --build-id <build-id>            Filter by the build ID of the source repository (default: "")
  -c, --commit-id <commit-id>          Filter by a commit ID from the source repository (default: "")
  -d, --deployment-id <deployment-id>  Filter by the deployment ID of the source repository (default: "")
  -i, --image-tag <image-tag>          Filter by a container image tag (default: "")
  -e, --env <environment>              Filter by environment name (default: "")
  -s, --service <service-name>         Filter by service name (default: "")
  -t, --top <top>                      Return only top N most recent deployments (default: "")
  -O, --output <output-format>         Output the information one of the following: normal, wide, JSON (default: "normal")
  -w, --watch                          Watch the deployments for a live view (default: false)
  -h, --help                           output usage information
```

andrebriggs: @dennisseah I agree with @samiyaakhtar. `-o` is universally an alias for `--output`or `--option`. I won't call it a standard because there isn't one but I'm willing to bet that `-o` alias for `--output` is intuitive to most (Unix style)

Also keeping the alias as lowercase in general reduces the mental overhead. See this [list](https://www.gnu.org/prep/standards/html_node/Option-Table.html#Option-Table) of options of GNU programs. The majority are lowercase but there is not specific pattern for when to use uppercase
. 
Again these are all opinions. Perhaps those with strong feelings can come up with a list of reserved short names

@samiyaakhtar @bnookala @evanlouie we have written guidelines [here](https://github.com/CatalystCode/spk/blob/master/guides/command-implementation.md#choosing-an-alias-for-option-name) that the team is free to change

dennisseah: first and foremost, upper and lower case alias should not be a problem. Being a solaris and linux users for over 20 years, I have never heard of any complaints. try `man sed` and you can see `-E` as alias. I wish to get this straight. Please feel free to disagree and this team can decide we can   only use lower case alias. 

secondly, this is what we have now
```
O: [
  "output",
  "output"
]
U: [
  "hld-repo-url"
]
_: [
  "commit-id",
  "default-component-git",
  "default-component-name",
  "default-component-path",
  "env",
  "git-push",
  "git-push",
  "helm-config-access-token-variable",
  "hld-commit-id",
  "image-tag",
  "k8s-backend",
  "k8s-backend-port",
  "manifest-commit-id",
  "middlewares",
  "p1",
  "p2",
  "p3",
  "path-prefix",
  "path-prefix-major-version",
  "pr",
  "remote-url",
  "repository",
  "service",
  "service-principal-id",
  "service-principal-password",
  "subscription-id",
  "target-branch",
  "tenant-id",
  "yaml-file-branch",
  "yaml-file-branch",
  "yaml-file-branch"
]
a: [
  "personal-access-token",
  "personal-access-token",
  "personal-access-token",
  "personal-access-token",
  "personal-access-token",
  "personal-access-token"
]
b: [
  "build-id",
  "build-script-url",
  "build-script-url",
  "build-script-url",
  "helm-config-branch"
]
c: [
  "commit-id",
  "helm-chart-chart"
]
d: [
  "deployment-id",
  "description",
  "devops-project",
  "devops-project",
  "devops-project",
  "devops-project",
  "devops-project",
  "packages-dir"
]
e: [
  "env",
  "maintainer-email"
]
f: [
  "file",
  "file",
  "file"
]
g: [
  "helm-config-git"
]
i: [
  "image-tag",
  "interactive"
]
k: [
  "access-key",
  "key-vault-name"
]
l: [
  "packages-dir",
  "storage-location"
]
m: [
  "maintainer-name",
  "manifest-url"
]
n: [
  "display-name",
  "name",
  "name",
  "pipeline-name",
  "pipeline-name",
  "pipeline-name"
]
o: [
  "org-name",
  "org-name",
  "org-name",
  "org-name",
  "org-name",
  "org-name"
]
p: [
  "helm-config-path",
  "partition-key",
  "port",
  "project",
  "service-principal-password"
]
r: [
  "default-ring",
  "helm-chart-repository",
  "registry-name",
  "remove-all",
  "storage-resource-group-name",
  "version"
]
s: [
  "self-test",
  "service",
  "source",
  "source-branch",
  "storage-account-name"
]
t: [
  "storage-table-name",
  "table-name",
  "template",
  "tenant",
  "title",
  "top"
]
u: [
  "hld-url",
  "repo-url",
  "repo-url",
  "service-principal-id"
]
w: [
  "watch"
]
```
`-o` is for `--org-name`. 
Hence we have `-O` for  `--output`.

`-u` is usually for `--user-id`, `service-provider-id`
Hence we have `-U` for all urls.

I find it very difficult to standardize the aliases because there were no attempts in the past.
@andrebriggs, please let me know how do we want to move forward with this. And I have written the guidelines and I do not know if people read it. All opinions and contributions are welcome.

andrebriggs: > I find it very difficult to standardize the aliases because there were no attempts in the past.

Thanks @dennisseah for attempting to drive some clarity. If folks really feel strongly about this please propose a plan. Let's avoid bikeshedding. 

The great thing is that the more generated documentation we have the easier it is to change these things in the future. Moreover, many of the commands will not be run by humans.

Let's wait until the end of Monday to see if anyone has a proposal otherwise lets commit (agreeable or disagreeably) and move on. 

cc @samiyaakhtar @bnookala @evanlouie

evanlouie: I agree with @samiyaakhtar and @andrebriggs that the short flag for `--output` should be a lowercase `-o` -- as the description and function of it aligns to what `kubectl` does (and since were operating in the realm of k8s, we should align to any standard they have). 
I recommend switching it with `--org-name` (or changing the short flag of `--org-name` to something else).

dennisseah: I have provided you the list of aliases used. can you please suggest what to use for --org-name?

bnookala: @dennisseah - I think we should consider the use cases for these commands, in which case I'm _not_ worried about `-o` being shadowed by something else. `spk deployment` commands do not shadow `-o` being azdo `org-name`, at least not yet… - and while you are right that we have a number of flags, with varying meanings, i do not see this as a reason to standardize, given the "roles" we've assumed for each of the commands, and flags ie `spk infra` associated with infrastructure, etc. `spk deployment` for the dashboard, and everything else for the pipelines/glue that holds it all together. 

as for what we can assume for this particular flag, i would like it to be left as `-o`, and for us to avoid utilizing upper case flags altogether.

dennisseah: I totally agree with you @bnookala. It is hard for me to do this form of housekeeping because I do not know these commands well enough. e.g. if with use -o for --output when there a chance that we need --org-name for the command. I am not sure. hence it is better for expert(s) for each command to fix them. We have 1. externalize the command decorator; 2. tool to view all aliases, `ts-node tools/locateAliases.ts`. It is clear that we need to do something to harmonize all the aliases; and we need to have aliases for options (all command line tools have them :-)). I will abandon this PR; and let member teams to fix them accordingly. Thanks

PR #370, opened 2020-03-05
Fix normalized name 

bnookala:
Fixes an issue in which normalized name was only replaced the first instance of an invalid character, instead of all instances of instances of the invalid character, and adds a test

### DIFF
--- a/src/commands/hld/reconcile.test.ts
+++ b/src/commands/hld/reconcile.test.ts
@@ -741,6 +741,15 @@ describe("normalizedName", () => {
     expect(normalizedName("fabrikam.frontend")).toBe("fabrikam-frontend");
   });
 
+  it("can handle multiple occurences of invalid characters", () => {
+    expect(normalizedName("fabrikam.frontend.foo")).toBe(
+      "fabrikam-frontend-foo"
+    );
+    expect(normalizedName("fabrikam/frontend/foo")).toBe(
+      "fabrikam-frontend-foo"
+    );
+  });
+
   it("can handle combinations of slashes and periods and caps in a name", () => {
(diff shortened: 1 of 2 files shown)


yradsmikham (approved):

src/commands/hld/reconcile.test.ts
```diff
@@ -741,6 +741,15 @@ describe("normalizedName", () => {
...
   });
 
+  it("can handle multiple occurences of invalid characters", () => {
+    expect(normalizedName("fabrikam.frontend.foo")).toBe(
+      "fabrikam-frontend-foo"
+    );
+    expect(normalizedName("fabrikam/frontend/foo")).toBe(
+      "fabrikam-frontend-foo"
+    );
+  });
```
dennisseah, line 751: would be nice to add a test case to test lower casing.

evanlouie (approved):

src/commands/hld/reconcile.ts
```diff
@@ -77,8 +77,8 @@ export interface IReconcileDependencies {
 export const normalizedName = (name: string): string => {
   return name
     .toLowerCase()
-    .replace("/", "-")
-    .replace(".", "-");
+    .replace(/\//g, "-")
+    .replace(/\./g, "-");
```
dennisseah, line 81: nitpicking: can be done with one pass
```
return name.toLowerCase().replace(/[/.]/g, "-");
```

src/commands/hld/reconcile.test.ts
bnookala: it already exists https://github.com/CatalystCode/spk/pull/370/files#diff-948e5cd512781080b25c861e85552153R732

PR #371, opened 2020-03-05
Add ring delete command

### LINKED ISSUE #971: Delete a ring in SPK
**As a:**
spk user
**I want:**
to be able to delete a ring
**So that:**
i can have remove a ring i no longer need
a new `spk ring delete <ring-name>` command to delete a ring from a spk managed project.

**Acceptance Criteria:**
- [ ] Create a new command `spk ring delete <ring-name>`
- [ ] Command removes the ring from the `bedrock.yaml`
- [ ] For all services linked in `bedrock.yaml`, remove the new ring from the trigger branches in the service's build pipeline
- [ ] This command will not allow the deletion of the default ring. Users must first change the default ring before deleting the current default ring. Setting default ring is captured here: https://github.com/microsoft/bedrock/issues/972

Yes

evanlouie:
Added the `spk ring delete <ring-name>` command.

closes https://github.com/microsoft/bedrock/issues/971

### DIFF
--- a/src/lib/bedrockYaml.test.ts
+++ b/src/lib/bedrockYaml.test.ts
@@ -1,4 +1,4 @@
-import uuid from "uuid/v4";
+import uuid = require("uuid/v4");
 import { createTempDir } from "../lib/ioUtil";
 import { createTestBedrockYaml } from "../test/mockFactory";
 import { IBedrockFile, IHelmConfig } from "../types";
@@ -10,7 +10,8 @@ import {
   fileInfo,
   isExists,
   read,
-  setDefaultRing
+  removeRing,
+  setDefaultRing,
 } from "./bedrockYaml";
 
 describe("Creation and Existence test on bedrock.yaml", () => {
@@ -76,11 +77,7 @@ describe("Adding a new service to a Bedrock file", () => {
     );
 
     const expected: IBedrockFile = {
-      rings: {
-        master: {
-          isDefault: true
-        }
-      },
+      ...defaultBedrockFileObject,
       services: {
         ...(defaultBedrockFileObject as IBedrockFile).services,
         ["./" + servicePath]: {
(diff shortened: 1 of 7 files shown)


src/lib/bedrockYaml.ts
```diff
@@ -151,3 +151,50 @@ export const fileInfo = (rootProjectPath?: string): IBedrockFileInfo => {
...
+ * Deletes the target ring with name `ringToDelete` from the provided `bedrock`
+ * config.
+ *
+ * @throws {Error} if ring is not found in `bedrock`
+ * @throws {Error} if the matching ring is `isDefault === true`
+ *
+ * @param bedrock the bedrock file to remove the ring from
+ * @param ringToDelete the name of the ring to remove
+ */
+export const removeRing = (
```
dennisseah, line 197: can we rename this to `deleteRing`?

evanlouie: I would want to keep `removeRing` so long as the function to add is still called `addNewRing`. Add/remove and create/delete are generally paired together as such.

src/lib/bedrockYaml.test.ts
```diff
@@ -1,4 +1,4 @@
-import uuid from "uuid/v4";
+import uuid = require("uuid/v4");
```
mtarng, line 1: Why the change from `from` to `require`?

mtarng (approved):

src/lib/bedrockYaml.test.ts
evanlouie: the `uuid` library uses a really outdated method for exporting its functions:

```ts
import { v4 } from './interfaces';

declare const v4: v4;

export = v4
```

the `export = v4` is non-standard and only worked with our old `import...from` call because our tsconfig has `"esModuleInterop": true`. However, since we've added `"exclude": ["**/*.test.ts"]` to our tsconfig as well, a lot of refactoring tools and linters flag ` import uuid from "uuid/v4";` as incorrect in tests files.

src/lib/bedrockYaml.ts
dennisseah: got it. thanks

dennisseah (approved):

PR #378, opened 2020-03-07
Refactor fab set to set image tag in service directory config file

### LINKED ISSUE #1100: fab set in build update hld pipeline should set image tag in different config file
**Describe the bug:**

As filed by this user issue: https://github.com/microsoft/bedrock/issues/1060, the `build-update-hld` pipeline will change the same common.yaml file for multiple service builds (either mono-repository or not mono repository)

**To Reproduce:**

1. have two services set up, building and configuring the same HLD repository
2. commit to one service, note the PR against the HLD - do not merge it.
3. commit to second service, second PR is generated against HLD
4. merge the first PR to the HLD
5. the second PR will not be able to be merged as there will be a conflict

**Expected behavior:**

both PRs should be able to be merged

**Additional context:**

this is likely happening as all PRs against the HLD update the same file. As there exists multiple places for configuring image tag etc, we should be able to set the configuration further down in the HLD - not just in the root level config/common.yaml - this should prevent any conflicts from two completely separate services updating the same configuration file. 

this line is where we set fab configuration :https://github.com/CatalystCode/spk/blob/master/src/lib/fileutils.ts#L301

yes

yradsmikham:
Close https://github.com/microsoft/bedrock/issues/1100.

Tested: https://dev.azure.com/yvradsmi/spk-services/_git/fabrikam-hld?path=%2Ffabrikam2019%2Ffabrikam-acme-frontend%2Fconfig%2Fcommon.yaml&version=GBDEPLOY%2Ffabrikam2019-fabrikam.acme.frontend-master-20200307.5&line=12&lineEnd=13&lineStartColumn=1&lineEndColumn=1&lineStyle=plain


andrebriggs (dismissed):
Seems like it wasn't tested if the `fileutils` tests is failing  :)

src/lib/fileutils.ts
```diff
@@ -298,7 +298,9 @@ export const serviceBuildAndUpdatePipeline = (
                   `echo "Image Name: $IMAGE_NAME"`,
                   `export IMAGE_REPO=${IMAGE_REPO}`,
                   `echo "Image Repository: $IMAGE_REPO"`,
-                  `../fab/fab set --subcomponent $(Build.Repository.Name).$FAB_SAFE_SERVICE_NAME.${SAFE_SOURCE_BRANCH}.chart image.tag=$IMAGE_TAG image.repository=$IMAGE_REPO/$BUILD_REPO_NAME`,
+                  `cd $(Build.Repository.Name)/$FAB_SAFE_SERVICE_NAME`,
```
bawcs: Is this working if I have 2 rings release at the same time?
What is the reason for not set image tag in ring directory ?

mtarng: I agree. We should probably set the config down at: `cd $APPLICATION_REPO_NAME/$FAB_SAFE_SERVICE_NAME/$RING_NAME` 

with:
`export APPLICATION_REPO_NAME=$(echo $(Build.Repository.Name) | tr '[:upper:]' '[:lower:]' | tr . - | tr / -)`,
and
`export RING_NAME=$(echo $(Build.SourceBranchName) | tr '[:upper:]' '[:lower:]' | tr . - | tr / -)`,

Though we need to double check what naming rules we have in place for rings. I think underscore's break it, so we may need to cover that case too.

@bnookala @evanlouie

yradsmikham: I raised this question with @bnookala, and wasn't sure if it needed to be at the ring level. Happy to make the changes if needed! :)

yradsmikham:
> Seems like it wasn't tested if the `fileutils` tests is failing :)

Resolved.

src/lib/fileutils.ts
bnookala: underscores do break - the name has to be dns compliant.

as for which level, let's do ring level as @luongbac301092 had suggested

yradsmikham: Sounds good! I'll make that change.

mtarng: some reference on the ring name needing to be massaged: https://github.com/microsoft/bedrock/issues/1103

bnookala (approved):

evanlouie (approved):

PR #379, opened 2020-03-07
[FEATURE] create service principal in spk setup

dennisseah:
related to https://github.com/microsoft/bedrock/issues/1113

in order to scaffold a sample app, we need a service principal.
user has 2 options
1. provide it (Id, password and Tenant Id)
2. get command line to execute az command line tool to create it.

### DIFF
--- a/src/lib/setup/servicePrincipalService.ts
+++ b/src/lib/setup/servicePrincipalService.ts
@@ -2,6 +2,11 @@ import { logger } from "../../logger";
 import { exec } from "../shell";
 import { IRequestContext } from "./constants";
 
+/**
+ * Login to az command line tool. This is done by
+ * doing a shell exec with `az login`; then browser opens
+ * prompting user to select the identity.
+ */
 export const azCLILogin = async () => {
   try {
     logger.info("attempting to login to az command line");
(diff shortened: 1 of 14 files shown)


andrebriggs (changes requested):
@dennisseah when manually entering the SP password the text is not hidden. Otherwise looks good!

dennisseah:
> @dennisseah when manually entering the SP password the text is not hidden. Otherwise looks good!

accept, thanks for testing it out

andrebriggs (approved):

src/lib/setup/servicePrincipalService.ts
```diff
@@ -0,0 +1,33 @@
...
+    logger.error(err);
+    throw err;
+  }
+};
+
+export const createWithAzCLI = async (rc: IRequestContext) => {
+  await azCLILogin();
+  try {
+    logger.info("attempting to create service principal with az command line");
+    const result = await exec("az", ["ad", "sp", "create-for-rbac"]);
```
sarath-p, line 34: Is there a SDK available to manage SP instead of using az cli?

As an example: https://docs.microsoft.com/en-us/azure/javascript/node-sdk-azure-authenticate-principal#create-a-service-principal-using-the-azure-sdk-for-nodejs

sarath-p (approved):

src/lib/setup/servicePrincipalService.ts
dennisseah: in order to user the SDK, we need to authenticate and hence we need a service principal. So we need to create a service principal or ask user to enter service principal id, password and tenantId. Once this is done, we can start to use the SDK.

PR #383, opened 2020-03-09
RFC1123 Compliant DNS Routes

### LINKED ISSUE #1103: Ring names must be DNS safe
**Describe the bug:**  IngressRoute names must be DNS compliant and match the regex `/[a-z-\.]+/` otherwise the manifest will fail to apply:

```
ts=2020-03-05T22:35:44.264996353Z caller=sync.go:536 component=cluster method=Sync cmd="kubectl apply -f -" took=450.661307ms err="running kubectl: The IngressRoute \"storage-service-msft_testing\" is invalid: metadata.name: Invalid value: \"storage-service-msft_testing\": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')" output=
```

**To Reproduce:**

bedrock.yaml:

```yaml
rings:
  msft_testing:
    isDefault: true
services:
  ./:
    displayName: storage-service
    helm:
      chart:
        branch: master
        git: 'https://some_git_repo.com'
        path: storage-service
    k8sBackend: chart-storage-service
    k8sBackendPort: 80
    middlewares: []
    pathPrefix: ''
    pathPrefixMajorVersion: ''
variableGroups:
  - msfttestingvg
```

IngressRoutes with the name `storage-service-msft_testing` will be created. These are invalid.

**Expected behavior:** The ingress route is given a valid DNS-1123 subdomain string as a name

**Screenshots:**

**Additional context:**

evanlouie:
- `spk hld reconcile` now generates RFC1123 compliant parameters for the
  generated Treafik IngressRoutes.
- `TraefikIngressRoute()` now throws when any combination of `ring`
  `serviceName` or `k8sBackend` generate a non RFC1123 compliant DNS name.

fixes https://github.com/microsoft/bedrock/issues/1103

### DIFF
--- a/src/lib/net/dns.ts
+++ b/src/lib/net/dns.ts
@@ -1,0 +1,61 @@
+////////////////////////////////////////////////////////////////////////////////
+// DNS helper functions
+//
+// Use these functions for sanitization and validation of DNS names and DNS name
+// segments.
+////////////////////////////////////////////////////////////////////////////////
+
+export const validDnsRegex = /^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$/;
+
+/**
+ * Returns if the dns string `segment` is composed only of DNS compliant
+ * characters.
+ * Passing this does not make the `segment` a valid DNS name -- use isValid()
+ * for that functionality.
+ *
+ * @param segment the string to validate contains only DNS compliant characters
+ */
+export const containsValidCharacters = (segment: string): boolean => {
+  return !!segment.match(/^[0-9a-z-.]+$/);
+};
+
+/**
+ * Returns if the `dns` string provided is a valid DNS-1123 name.
+ *
+ * @param dns string to validate is a DNS compliant string
+ */
+export const isValid = (dns: string): boolean => {
+  return !!dns.match(validDnsRegex);
+};
+
+/**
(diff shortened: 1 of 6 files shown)


src/lib/net/dns.ts
```diff
@@ -0,0 +1,61 @@
+////////////////////////////////////////////////////////////////////////////////
+// DNS helper functions
+//
+// Use these functions for sanitization and validation of DNS names and DNS name
+// segments.
+////////////////////////////////////////////////////////////////////////////////
+
+export const validDnsRegex = /^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$/;
```
dennisseah, line 8: Do we need to check for uppercase characters? A-Z

evanlouie: Although RFC1123 DNS spec says DNS names are case sensitive, k8s DNS only supports lower case (that regex is actually ripped straight from k8s DNS core validator)

dennisseah: ok

dennisseah (approved):

bnookala (approved):

PR #386, opened 2020-03-10
[FEATURE] Get subscription id in spk setup command

### LINKED ISSUE #1138: Acquiring Subscription Id in spk setup command
**As a:**
user of spk setup command

**I want:**
the command to figure out the subscription Id automatically

**So that:**
I do not need to enter it.

Using the service principal, we can figure out the subscription Id.

documented in the command md file

Related to https://github.com/microsoft/bedrock/issues/1113

dennisseah:
In the event that there are more than one subscription, user needs to select one of them. Display names shall be displayed as choices.

the subscription id value will be written to spk config.yaml file

closes https://github.com/microsoft/bedrock/issues/1138


src/lib/setup/subscriptionService.ts
```diff
@@ -0,0 +1,45 @@
...
+      .then(async (creds: ApplicationTokenCredentials) => {
+        const client = new SubscriptionClient(creds);
+        const subsciptions = await client.subscriptions.list();
+        const result = (subsciptions || []).map(s => {
+          return {
+            id: s.subscriptionId!,
+            name: s.displayName!
+          };
+        });
+        logger.info("Successfully acquored subscription list");
```
sarath-p: typo `acquored`

src/lib/setup/subscriptionService.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/**
+ * Returns a list of subscriptions based on the service principal credentials.
+ *
+ * @param rc Request Context
+ */
+export const getSubscriptions = (
+  rc: IRequestContext
+): Promise<ISubscriptionItem[]> => {
```
sarath-p, line 21: why not using `async` since the entire code base using it? `Promise` is used in few places due to SDK dependency.

src/lib/setup/prompt.ts
```diff
@@ -1,15 +1,41 @@
...
+      {
+        choices: subscriptions.map(s => s.name),
+        message: "Select one of the subscription\n",
+        name: "az_subscription",
+        type: "list"
+      }
+    ];
+    const ans = await inquirer.prompt(questions);
+    rc.subscriptionId = subscriptions.find(
+      s => s.name === ans.az_subscription
```
sarath-p, line 35: I think adding a case insensitive would be good when checking for subscription names.

sarath-p (changes requested):

andrebriggs (approved):
Look good overall. Pending the comments from others

src/lib/setup/subscriptionService.ts
dennisseah: accept

src/lib/setup/subscriptionService.ts
dennisseah: can you show me how you want the code to be written? async keyword is needed when we have await in the function; and wrapping the entire function as a Promise

src/lib/setup/prompt.ts
dennisseah: this is not necessary. because the name is returned by SDK; and we are formulating the question based on these names and then match them with the id. Can you tell me why do not need to case insensitive check here?

sarath-p: If user is selecting the name from a list of options, it is fine. I missed that part earlier. If user enters name, I would make case insensitive match in this scenario.

src/lib/setup/subscriptionService.ts
sarath-p: it is already returning `promise` which means you are using asynchronous or blocking code in the function. `async` functions use an implicit Promise to return results anyway. Keeping code consistent would be good.

dennisseah: my point is we do not add async unnecessary and it is not encourage. it is not about consistency.
Sorry I do not get your point on why you want to add async?

dennisseah: I can write the function in the manner but not simply adding an async keyword.
```
export const getSubscriptions = async (
  rc: IRequestContext
): Promise<ISubscriptionItem[]> => {
  logger.info("attempting to get subscription list");
  const creds = await loginWithServicePrincipalSecret(
    rc.servicePrincipalId!,
    rc.servicePrincipalPassword!,
    rc.servicePrincipalTenantId!
  );
  const client = new SubscriptionClient(creds);
  const subsciptions = await client.subscriptions.list();
  const result = (subsciptions || []).map(s => {
    return {
      id: s.subscriptionId!,
      name: s.displayName!
    };
  });
  logger.info("Successfully acquired subscription list");
  return result;
};
```

sarath-p (approved):

PR #392, opened 2020-03-11
Exit on invalid DNS user input

### LINKED ISSUE #1128: spk service create and spk ring create should error and exit when non dns compliant names are passed as parameters
**As a:**

spk user and developer

**I want:**

`spk service create` and `spk ring create` to error and exit on bad inputs

**So that:**

I can catch potential errors at service creation time rather than when my resources are generated and applied to a clsuter

`spk service create` and `spk ring create` to error and exit when the following arguments are _not_ dns compliant

spk service create:
- `-n` or `--display-name` (for service create, also used in the name of generated ingress routes)
- `--k8s-backend` (the internal k8s service name, also used in internal service dns name, and mapping service to ingress route)
- `--path-prefix` (path prefix on the external facing ingress route)
- `--path-prefix-major-version` (major version string that prefixes the path prefix)

spk ring create/add:
- ring name (used for ingress route name and ingress route header matching 

**Acceptance Criteria:**

- [x] `spk service create --display-name` should _only_ accept a DNS compliant name
- [x] `spk service create --k8s-backend` should _only_ accept a DNS compliant name
- [x] `spk service create --path-prefix` should _only accept a DNS compliant name
- [x] `spk service create --path-prefix-major-version `should _only_ accept a DNS compliant name
- [x] `spk ring create` should only accept a DNS compliant ring name

**Additional context:**

k8s service names have to be DNS compliant, as do traefik rules - as we use a combination of parameters to generate the match rules and backing service names, we should try and catch any potential errors from bad inputs at the earliest point we can

see: https://github.com/CatalystCode/spk/blob/master/guides/building-helm-charts-for-spk.md#static-configuration for more information on how routes are generated from config in bedrock.yaml

DNS name validation has been added, and can be utilized as helper functions: https://github.com/CatalystCode/spk/pull/383/files

yes

evanlouie:
- `spk service create` and `spk ring create` now log an error and exit when any
  arguments are inputted which are not RFC1123 compliant.
  - The checks only occur when the input has been explicitly passed by the user
    (i.e. not `undefined` or `""`)

fixes https://github.com/microsoft/bedrock/issues/1128

### DIFF
--- a/src/commands/service/create.ts
+++ b/src/commands/service/create.ts
@@ -21,6 +21,7 @@ import {
   generateServiceBuildAndUpdatePipelineYaml
 } from "../../lib/fileutils";
 import { checkoutCommitPushCreatePRLink } from "../../lib/gitutils";
+import * as dns from "../../lib/net/dns";
 import { isPortNumberString } from "../../lib/validator";
 import { logger } from "../../logger";
 import { IBedrockFileInfo, IHelmConfig, IUser } from "../../types";
@@ -54,7 +55,7 @@ export interface ICommandValues extends ICommandOptions {
 
 export const fetchValues = (opts: ICommandOptions) => {
   if (!isPortNumberString(opts.k8sBackendPort)) {
-    throw new Error("value for --k8s-service-port is not a valid port number");
+    throw Error("value for --k8s-service-port is not a valid port number");
   }
 
   const bedrock = Bedrock();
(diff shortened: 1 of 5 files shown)


src/commands/service/create.ts
```diff
@@ -116,12 +142,20 @@ export const execute = async (
...
     return;
   }
 
+  // validate user inputs are DNS compliant
+  try {
+    assertValidDnsInputs(opts);
+  } catch (err) {
+    logger.error(err);
+    await exitFn(1);
+  }
```
dennisseah, line 157: please add a `return;` after the `await exitFn(1);`. otherwise the code after this will continue to run because there is a time gap before `process.exit(1)` is executed. This is due to the wait time for winston.logger to flush buffer to log file

dennisseah: take it back. it should be ok.

dennisseah (approved):

mtarng (approved):

mtarng:
lint error @evanlouie

yradsmikham (approved):

PR #395, opened 2020-03-12
Migrate from TSLint to ESlint

### LINKED ISSUE #1157: Migrate from tslint to eslint
**As a:** SPK developer

**I want:** to use eslint instead of tslint for linting

**So that:** we are no longer using a deprecated linting library/tool

- remove tslint library from code base as well as the `tslint.json` file
- follow this guide for setting up eslint with typescript support: <https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/README.md>
- fix all new linting errors which show up due to the change in linter

**Acceptance Criteria:**

Refer to above solution

**Describe alternatives you've considered:** N/A

**Additional context:** 

- When we initially setup SPK, tslint had just recently been deprecated and eslint with typescript support was still in its infancy and fairly buggy. It has reached a point of maturity that it is now recommend to use instead of tslint
- Because eslint is more powerful than tslint, after the change in linter, many linting errors will show. This will require quite a bit of code change to fix. These will most likely be all cosmetic. (ie; function/variable namings, order of function definitions, etc...)

evanlouie:
- Removed TSLint and tslint.json configuration file
- Added ESLint with `recommended` TypeScript configurations.
- Removed all `I` prefix from type declarations (per ESLint recommendation)
- Have added return types to all function declarations (per ESLint
  recommendation)
- To have minimal conflict with integration, many of the recommended ESLint
  rules have been disabled on a per-page/line basis.
  - We should attempt to remove these `/* eslint-disable ...` lines as refactor
    code and conform completely to the recommended linting rules.

closes https://github.com/microsoft/bedrock/issues/1157


src/lib/fileutils.ts
```diff
@@ -595,7 +599,7 @@ const manifestGenerationPipelineYaml = () => {
           `download_spk`,
           `message="$(Build.SourceVersionMessage)"`,
           `if [[ $message == *"Merged PR"* ]]; then`,
-          `pr_id=$(echo $message | grep -oE '[0-9]+' | head -1 | sed -e 's/^0\+//')`,
+          `pr_id=$(echo $message | grep -oE '[0-9]+' | head -1 | sed -e 's/^0\\+//')`,
```
evanlouie, line 602: @samiyaakhtar Was I correct in adding this additional `\`? 
without it, the `\` acts as a noop and the outputted string becomes `pr_id=$(echo $message | grep -oE '[0-9]+' | head -1 | sed -e 's/^0+//')`

samiyaakhtar: @evanlouie Looks good to me

samiyaakhtar (approved):

dennisseah (approved):
just manage to briefly go thru the changes. too much to review :-)

yradsmikham (approved):

PR #402, opened 2020-03-16
Changes to get build pipeline working in branch issue1113x

andrebriggs:


dennisseah (approved):

src/lib/azure/containerRegistryService.ts
```diff
@@ -158,15 +157,14 @@ export const create = async (
...
     name
   );
 
   if (exist) {
     logger.info(
-      `Azure container registry, ${name} in ${resourceGroup} already existed`
+      `Azure container registry, ${name} already existed in subscription`
     );
-    return false;
+    return true;
```
dennisseah, line 167: this should be false because we want to indicate that the command does not create the ACR in the setup.log

PR #404, opened 2020-03-16
Remove ./ From Path Triggers

### LINKED ISSUE #1168: Azure Pipelines path triggers cannot begin with "."
**Describe the bug:**

When using a mono-repository, for managing services, there exist a path in the azure-pipelines.yaml path trigger that adds a period (".") to the path prefix. This is causing pipelines to _not_ run. 

**To Reproduce:**

Add a service to a mono-repository, and note that the azure-pipelines.yaml for the build-update-hld.yaml file for that service has a period prefixed to it. 

**Expected behavior:**

Pipelines should run

**Screenshots:**

**Additional context:**

This seems to be new behavior in AzDo

evanlouie:
- Path triggers for service build update pipelines are now sanitized to not
  include any leading `./` in the path.
- Removed old `tslint` flags.

fixes https://github.com/microsoft/bedrock/issues/1168

### DIFF
--- a/src/lib/fileutils.ts
+++ b/src/lib/fileutils.ts
@@ -162,6 +162,18 @@ export const generateServiceBuildAndUpdatePipelineYaml = (
 };
 
 /**
+ * Sanitize the given path to format Azure DevOps can properly utilize
+ *
+ * Transforms:
+ * - If present, removes leading dot-slash (`./`) prefix from the path
+ *
+ * @param pathLike a path-like string to sanitize
+ */
+export const sanitizeTriggerPath = (pathLike: string): string => {
+  return pathLike.replace(/^\.\//, "");
+};
+
+/**
  * Returns a build-update-hld-pipeline.yaml string
  * based on: https://github.com/andrebriggs/monorepo-example/blob/master/service-A/azure-pipelines.yml
  *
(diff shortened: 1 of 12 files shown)


src/lib/fileutils.ts
```diff
@@ -161,6 +161,18 @@ export const generateServiceBuildAndUpdatePipelineYaml = (
...
+/**
+ * Sanitize the given path to format Azure DevOps can properly utilize
+ *
+ * Transforms:
+ * - If present, removes leading dot-slash (`./`) prefix from the path
+ *
+ * @param pathLike a path-like string to sanitize
+ */
+export const sanitizeTriggerPath = (pathLike: string): string => {
+  return pathLike.replace(new RegExp("^./"), "");
```
dennisseah: i think it should be
```
return pathLike.replace(/^\.\//, "");
```
because `.` means anything in regular expression hence it needs to be escaped.

would be nice if you can add `sanitizeTriggerPath("a/bar")` in your test.

evanlouie: woops. Forgot what i needed to escape since I was using `new RegExp` instead of `//`. Changed and added another test.

dennisseah (approved):

mtarng (approved):

PR #415, opened 2020-03-18
[FEATURE] Supporting introspection configuration setup in spk init -interactive command

### LINKED ISSUE #1174: support spk deployment resources in spk init command
**As a:**
user

**I want:**
to have spk deployment resources configured for me in an interaction mode

**So that:**
I do not need to look up the documentation on how to do this.

using `spk init` command to configure them.

**Acceptance Criteria:**

**Additional context:**

Yes, in the md file of `spk init` command

related to https://github.com/microsoft/bedrock/issues/1127

dennisseah:
closes https://github.com/microsoft/bedrock/issues/1174

based on design doc https://github.com/CatalystCode/spk/blob/master/technical-docs/designs/initialization/interactiveModeForIntrospectionConfig.md

Note: we are sharing code between `spk setup` and `skp init` commands so that there are some refactoring done to enable sharing of code.


andrebriggs (approved):

src/commands/init.ts
```diff
@@ -134,20 +114,79 @@ export const validatePersonalAccessToken = async (
...
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const azure = curConfig.introspection!.azure!;
+
+  const ans = await inquirer.prompt([
+    promptBuilder.azureStorageAccountName(azure.account_name),
+    promptBuilder.azureStorageTableName(azure.table_name),
+    promptBuilder.azureStoragePartitionKey(azure.partition_key),
+    promptBuilder.azureStorageRepoAccessKey(azure.source_repo_access_token),
```
samiyaakhtar: `azureStorageRepoAccessKey` should not be associated with `azure.source_repo_access_token`? Those are two different fields

src/commands/init.ts
```diff
@@ -134,20 +114,79 @@ export const validatePersonalAccessToken = async (
...
+    promptBuilder.azureKeyVaultName(curConfig.key_vault_name)
+  ]);
+  azure["account_name"] = ans.azdo_storage_account_name;
+  azure["table_name"] = ans.azdo_storage_table_name;
+  azure["partition_key"] = ans.azdo_storage_partition_key;
+  azure["source_repo_access_token"] = ans.azdo_storage_repo_access_key;
```
samiyaakhtar: for now, we probably don't need to prompt for source_repo_access_token since this is more for github repos, we can use the access key from azure devops section for that. So same goes here, `source_repo_access_token` should not be associated with `azdo_storage_repo_access_key` since those are different fields

src/lib/promptBuilder.ts
```diff
@@ -0,0 +1,181 @@
...
+};
+
+export const azureStorageRepoAccessKey = (
+  defaultValue?: string | undefined
+): QuestionCollection => {
+  return {
+    default: defaultValue,
+    mask: "*",
+    message: `${i18n.prompt.storageRepoAccessKey}\n`,
+    name: "azdo_storage_repo_access_key",
```
samiyaakhtar: perhaps we should call it `azdo_storage_access_key` (remove `repo` from it) since it is not associated with a repo

src/lib/validator.ts
```diff
@@ -260,3 +332,25 @@ export const validateACRName = (value: string): string | boolean => {
   }
   return true;
 };
+
+export const validateStorageKeVaultName = (value: string): string | boolean => {
```
samiyaakhtar: nit: `validateStorageKeVaultName` is missing a `y`, should be `validateStorageKeyVaultName`

samiyaakhtar (changes requested):

src/commands/init.ts
dennisseah: accept
I will rename to helper function to avoid confusion. `azureStorageRepoAccessKey` is just a helper function. 
one is helper function and another is configuration variable.

src/commands/init.ts
dennisseah: accept
just make it clear the `source_repo_access_token` is just a key from prompt answer.
I will remove the prompt for `source_repo_access_token`

src/lib/validator.ts

src/lib/promptBuilder.ts
dennisseah: remove this function base on your feedback.

samiyaakhtar (approved):
Was able to run `spk deployment get` after initializing with --interactive. Thanks Dennis!

PR #427, opened 2020-03-20
Default Routing for isDefault Ring

### LINKED ISSUE #1084: Fallback/Default routing for rings
**As a:** operator

**I want:** requests made to a service made through the Traefik2 ingress to have a fallback/default ring to route to when my request does not have `Ring` header.

**So that:**

- The on-boarding for a _ring-less_ spk project is easier and the concept of rings can be adopted when needed.
- In the case of a mono-repository with multiple services and rings, it is possible for a a ringed-service to call another ringed-serviced. But in the scenario that this ring is a one-off bug fix ring, chances are the other service has not been deployed to that ring. In that scenario, the request should be routed to a sane default (e.g `master).

- Only allow one entry in `rings` in `bedrock.yaml` to be marked as `isDefault: true`.
- During IngressRoute creation time, use that `isDefault: true` property as a trigger to scaffold an additional IngressRoute -- one without the header requirement.

**Acceptance Criteria:**
- [ ] Documentation in SPK guides describing this behavior 
- [x] Add to the `spk hld reconcile` functionality to create an additional IngressRoute if the a ring is flagged as `isDefault: true` -- the Ingress route should be identical to the other one except without the `Ring` header match rule.
- [x] Add a validation in IngressRoute creation time that only one ring can be flagged as `isDefault: true`.

**Additional context:**

This would impact the create-revision flow as it currently allow users to make PRs against the multiple rings defined with isDefault: true. However, this functionality has seen little real world use, so co-opting it might be best.

**Does this require updates to documentation?:** Potentially -- if the rings documentation is merged by the time this done.

evanlouie:
- When a ring is marked as `isDefault: true`, `spk hld reconcile` will now
  generate an additional IngressRoute and Middleware.
  - The IngressRoute will not have a `Ring` route match rule
  - The IngressRoute points to the same backend service as its ringed
    counterpart.
  - Only one ring can be marked `isDefault: true` -- validation is run at
    `spk hld reconcile` execute time, throwing an error if more than one
    `isDefault`.
- Refactored IngressRoute tests
- Added compatibility configuration for eslint and prettier.

closes https://github.com/microsoft/bedrock/issues/1084

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,4 +1,5 @@
 module.exports = {
-  preset: 'ts-jest',
-  testEnvironment: 'node',
+  preset: "ts-jest",
+  testEnvironment: "node",
+  rootDir: "src"
 };
--- a/src/commands/hld/reconcile-unit.test.ts
+++ b/src/commands/hld/reconcile-unit.test.ts
@@ -1,0 +1,151 @@
+import * as fs from "fs";
+import * as mocks from "../../test/mockFactory";
+import { BedrockFile } from "../../types";
+import * as reconcile from "./reconcile";
+
+jest.mock("fs");
+
+beforeEach(() => {
+  jest.resetAllMocks();
+});
+
+describe("createMiddlewareForRing", () => {
+  const tests: {
+    name: string;
+    actual: () => unknown;
+    expected: unknown;
+    effects?: (() => void)[];
+  }[] = [
+    {
+      name: "isDefault === false; creates one (1) middleware",
+      actual: (): unknown =>
+        reconcile.createMiddlewareForRing(
+          "path-to-ring",
+          "my-service",
+          "my-ring",
+          "v1",
+          false
+        ),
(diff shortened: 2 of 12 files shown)


evanlouie:
Investigating why tests are failing. Can't seem to make them fail locally.

edaena (approved):

mtarng:
Trying to follow the logic @evanlouie , what happens when we changed the default ring? will `spk hld reconcile` overwrite the existing IngressRoute definition in the hld repository?

evanlouie:
@mtarng yup. `reconcile` regenerates the IngressRoute and middlewares file every time. If a ring is changed from `isDefault: true` to `isDefault: false`, the regenerated ingressroute/middleware files will only have the ringed versions in it (as opposed to the multi docs (`---` delimitted) yaml files it outputs if it is `isDefault`)

mtarng (approved):

PR #432, opened 2020-03-22
[HOUSEKEEPING] Finding broken links in md files and fix them

### LINKED ISSUE #1214: Broken Links in md files
**Describe the bug:**
there are broken links in several md files under guides folder

**To Reproduce:**
e.g. go to https://github.com/CatalystCode/spk/blob/master/guides/service-introspection-onboarding.md and click on here link at "More information about the commands is available in the command reference __here__."

**Expected behavior:**
md files to have workable links

**Screenshots:**

**Additional context:**

No

dennisseah:
1. created a simple tool to find broken links. `ts-node tools/brokenLinks.ts`
2. fixed some eslint issues in other tools
3. fixed the broken links. Cannot figure out this one though. https://github.com/CatalystCode/spk/blob/master/guides/infra/spk-terragrunt-overview.md, we have a reference to https://github.com/CatalystCode/spk/blob/master/SPK-Infra.md

closes https://github.com/microsoft/bedrock/issues/1214


andrebriggs (approved):
Thank you @dennisseah for doing this

tools/brokenLinks.ts
```diff
@@ -0,0 +1,80 @@
+// DISCLAIMER: Only test this tool on mac
+// There may be some false negative results
```
yradsmikham: Why are we building a personal tool to check for broken links as opposed to using some kind of text or markdown linting?

And why does it only work on mac?

dennisseah: it is NOT a personal tool, it is for all the people who have linux or macOS. 
it work for mac because I do not test on linux and I cannot claim that it works on linux :-)
it does not work for windows because in markdown `/` is used for path. we have to replace them with '\' inorder for windows OS to work.

Please let me know if you think that this tool is not useful. I will not have it in git if your answer is yes.

thanks

yradsmikham: Sorry, by "personal",  I head meant a tool made specifically used for SPK codebase. I am fine with keeping it. But again, should we be looking into a text/markdown lint tool for our repo?

dennisseah: i cannot find any lint tools for markdown hence I create this one (just to check for broken links)

samiyaakhtar: @dennisseah how does this tool alert us about future broken links?

samiyaakhtar: I guess my question is shouldn't we be running this as part of the CI, so that we see a build failure when there's a broken link?

dennisseah: i do not know how can we do this. please provide suggestion. I have just written a simple tool to check for broken links and fix them. Please check with Andre if you need something more.

bnookala:
I did a bit of searching and found a few existing packages/tools that do similar tasks - I would err on the side of using them.

https://github.com/tcort/markdown-link-check - this one integrates with github actions, so can be run in CI as well: https://github.com/marketplace/actions/markdown-link-check

There's also this plugin for remark: https://github.com/remarkjs/remark-validate-links

bnookala (changes requested):
let's try to use this tool instead, since it's being actively maintained and might fit our use cases: https://github.com/tcort/markdown-link-check

dennisseah:
> let's try to use this tool instead, since it's being actively maintained and might fit our use cases: https://github.com/tcort/markdown-link-check

ok. I will remove my tool. thanks

mtarng (approved):

yradsmikham (approved):

tools/mdLint.ts
```diff
@@ -0,0 +1,38 @@
...
+const testMd = async (f: string): Promise<void> => {
+  try {
+    await exec("./node_modules/.bin/markdown-link-check", [f]);
+  } catch (e) {
+    console.log(`${f}: ${e.message}`);
+    process.exit(1);
+  }
+};
+
+(async (): Promise<void> => {
```
bnookala: I think this script file may be un-necessary:

See: https://github.com/tcort/markdown-link-check#check-links-from-a-local-markdown-folder-recursive 

You can inline this into a package.json script instead of calling this custom script

bnookala (changes requested):

tools/mdLint.ts
dennisseah: i cannot because i need a status code 1 to stop the script. :-)
the inline approach always return 0

evanlouie: Sorry Dennis, not quite sure I understand what you're saying. Npm scripts will exit with 1 if if any of the commands fail:

```json
  "scripts": {
    "i-will-fail": "false && true"
  },
```

```sh
npm run i-will-fail # will exit with code 1
```

are you thinking othewise?

dennisseah: maybe you can try it yourself and let me know what you get
add this to your package.json
```
  "md-lint": "find guides -name \\*.md -exec markdown-link-check {} \\",
```
or just ran (at the root of spk repo)
```
find guides -name \*.md -exec markdown-link-check {} \;
```
then 
```
echo $?
```
there are errors reported by the tool and it is returning 0 status code.

bnookala: @dennisseah try the suggestion here: https://github.com/tcort/markdown-link-check/issues/57#issuecomment-407061525

```
find guides -name \*.md | xargs -n 1 markdown-link-check
```

dennisseah: @bnookala 
and what about people using windows?

bnookala: @dennisseah this isn't a run-time requirement - just a development requirement. I expect that developers who are running windows who also want to contribute to spk development should do their work in a WSL environment.

dennisseah: ok then. We do not support windows runtime.

bnookala (approved):

PR #434, opened 2020-03-23
[HOUSEKEEPING] remove some eslint disablements

dennisseah:
1. remove the eslint-disable for the entire file so that eslint error will not be masked for new code.
2. reorder functions to address `no-use-before-define` error


dennisseah:
abandon this PR until https://github.com/CatalystCode/spk/pull/433 is merged

PR #443, opened 2020-03-24
[DOC-FIX] remove CORS setting requirement

dennisseah:


edaena:
@dennisseah Why is this requirement removed?

dennisseah:
> @dennisseah Why is this requirement removed?

We implemented the backend for dashboard and CORS settings are not needed. @samiyaakhtar can you confirm. Generally setting CORS for localhost is not good things (w.r.t. security)

samiyaakhtar:
@dennisseah @edaena shouldn't be needed anymore if people are using the new docker image hopefully, that queries APIs from a backend instead. I guess we can remove it since we've already updated the image

samiyaakhtar (approved):

edaena:
@samiyaakhtar @dennisseah do we need to also remove enabling CORS in the storage account in the integration tests?

andrebriggs (approved):

samiyaakhtar:
@edaena Agreed, maybe we should get rid of that snippet of code from integration tests as well. Might be worth closing out in one PR. @dennisseah could we remove the part from rings integration test which adds CORS setting to the storage account?

dennisseah:
> @edaena Agreed, maybe we should get rid of that snippet of code from integration tests as well. Might be worth closing out in one PR. @dennisseah could we remove the part from rings integration test which adds CORS setting to the storage account?

can someone do it? or tell me where is it if you want me to remove it :-)

samiyaakhtar:
@dennisseah it had to be done before the PR was merged :( I'll open another PR for this.

PR #445, opened 2020-03-24
added logging for generating config.yaml using spk init

### LINKED ISSUE #1226: Specifying an config file as a parameter for spk init creates a copy of the file resulting in 2 config files (original + config.yaml)
Feedback:
> In ~/.spk I now have 2 config.yaml files that are exact duplicates, given that spk-config.yaml was passed as a paramter the user is unaware of the other file and this leads to confusion which file to maintain going forward.   Consider providing a message back to the user that an "import" of the file was made and that going forward the config file is located at ~/.spk/config.yaml.

yradsmikham:
Close https://github.com/microsoft/bedrock/issues/1226.


src/config.ts
```diff
@@ -266,11 +266,13 @@ export const saveConfiguration = (
   targetDir: string = defaultConfigDir()
 ): void => {
   try {
+    logger.info("Generating config.yaml from yaml file input.");
     const data = yaml.safeDump(readYaml<ConfigYaml>(sourceFilePath), {
       lineWidth: Number.MAX_SAFE_INTEGER,
     });
     const targetFile = path.join(targetDir, "config.yaml");
     fs.writeFileSync(targetFile, data);
+    logger.info("config.yaml was generated and located at ~/.spk/config.yaml.");
```
mtarng: Let's log the value of `targetFile` instead of `~/.spk/config.yaml` just to be accurate.

mtarng (changes requested):

src/config.ts
yradsmikham: Done

yradsmikham: Done

sarath-p (approved):
LGTM

dennisseah (approved):

samiyaakhtar (approved):

mtarng (approved):

PR #446, opened 2020-03-24
Handle exception - first implementation

dennisseah:
test implementation on infra scaffold command. please provide feedback.


src/lib/errorStatusCode.ts
```diff
@@ -0,0 +1,14 @@
+// please do not change the status code numbers
+// you can add new ones but not changing the existing ones
```
andrebriggs, line 2: @dennisseah I assume the guidance is to add increasing error status codes with spaces in between to give ourselves room to add related error statuses?

andrebriggs (approved):

andrebriggs:
@dennisseah it might help to expand on [this](https://github.com/CatalystCode/spk/blob/master/guides/command-implementation.md#error-messages) section with the concrete implementation we have here. This way we can codify this pattern

src/lib/errorStatusCode.ts
dennisseah: precisely

dennisseah:
> @dennisseah it might help to expand on [this](https://github.com/CatalystCode/spk/blob/master/guides/command-implementation.md#error-messages) section with the concrete implementation we have here. This way we can codify this pattern

I would have to update my design doc once I have collected all feedbacks on this PR and then I will update that section.  hope that this is ok with you.

src/lib/errorStatusCode.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const errorStatusCode = {
+  1000: "command fails to execute",
```
edaena: nit: the numbers can also be mapped to an enum or a constat, VALIDATION_ERROR, EXECUTION_ERROR, ENVIRONMENT_ERROR, GIT_ERROR, COMMAND_FAIL_ERROR, or something similar

dennisseah: accept. will make the change. thanks

NathanielRose (approved):
Unit and Int tests work. LGTM

yradsmikham (approved):

edaena (approved):

PR #449, opened 2020-03-25
fix eslint warnings/errors in Deployment Dashboard command

dennisseah:
related to microsoft/bedrock#1157

after switching to eslint, we have disabled some rules. This is to remove eslint-disable in deployment dashboard command by using an interface to gather all the values needed for the command. e.g. avoid using non null assertion operations like this `config.introspection!.azure!.source_repo_access_token`

### DIFF
--- a/src/commands/deployment/dashboard.test.ts
+++ b/src/commands/deployment/dashboard.test.ts
@@ -11,6 +11,7 @@ import {
   logger,
 } from "../../logger";
 import {
+  DashboardConfig,
   execute,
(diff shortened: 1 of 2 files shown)


evanlouie (approved):

dennisseah:
@samiyaakhtar I believe https://github.com/microsoft/bedrock/issues/1244 is targeting at spektate repo and this PR is on spk repo. Thanks

samiyaakhtar (approved):

PR #455, opened 2020-03-26
[HOUSEKEEPING] remove eslint-disable for hld reconicle command

dennisseah:
mainly moving functions around to satisfy `no-use-before-define` rule
no changes in logics in this file

related to https://github.com/microsoft/bedrock/issues/1157


andrebriggs (approved):
Just noticing how long `reconcile.ts` is 😅

evanlouie (approved):

bnookala (commented):
@dennisseah In the future, I would like for us to track this sort of housekeeping work on the sprint board - please make sure you are linking an issue/opening an issue so that we know it is happening.

bnookala (changes requested):
@dennisseah Please add an issue to github, and add it to the sprint, as that's our process.

dennisseah:
> @dennisseah In the future, I would like for us to track this sort of housekeeping work on the sprint board - please make sure you are linking an issue/opening an issue so that we know it is happening.

sorry I forgot to link to an issue. I checked with Evan about removing the `eslint-disabled` when I have time. And he gave the green light. We have a lot of these after we switch from `tslint` to `eslint`

PR #458, opened 2020-03-26
[HOUSEKEEPING] Remove/Comment Out resources related to key vault for …

### LINKED ISSUE #1243: Remove the use of keyvault for spk deployment onboard from docs and command
**As a:** developer

**I want:** to understand the use of keyvault

**So that:** I can onboard onto using bedrock and service introspection easily 

There's some dangling code in the spk repo that is no longer being used, since it was originally added to allow users to upload their secrets onto keyvault when they used the `spk deployment onboard` command. It stores the storage access key in keyvault. However, with some recent upgrades to onboarding and with the help of this new setup command, the link that tied up keyvault with the storage access key config has been removed. 
- [x] Discuss the possibility of keeping this keyvault feature and bring back the links that were removed
- [ ] Otherwise, remove the code from displaying in the `spk` command and documentation but keep the business logic and tests. 

**Acceptance Criteria:**
Code is cleaned up

**Additional context:**

dennisseah:
…introspection

closes https://github.com/microsoft/bedrock/issues/1243

@samiyaakhtar, Please let me know if it is ok to comment out code; or you want to remove them. thanks


src/commands/deployment/onboard.decorator.json
```diff
@@ -22,10 +22,6 @@
       "description": "Name of the resource group to create new storage account when it does not exist",
       "required": true
     },
-    {
-      "arg": "-k, --key-vault-name <key-vault-name>",
-      "description": "Name of the Azure key vault"
-    },
     {
       "arg": "--service-principal-id <service-principal-id>",
```
samiyaakhtar, line 26: Are `service-principal-id`, `service-principal-password`, `tenant-id` and `subscription-id` still being used elsewhere?

samiyaakhtar:
Personally, I prefer removing lines of code that link to things that we don't plan to use since it appears like a user error if things are commented out, but we can leave the keyvault files as is. Does anyone have a preference on how to unlink some code for the moment until we need keyvault again? cc @sarath-p @evanlouie @bnookala

src/commands/deployment/onboard.decorator.json
dennisseah: in the onboard command these are used to create account service and table. Isn't this the case?

dennisseah:
> Personally, I prefer removing lines of code that link to things that we don't plan to use since it appears like a user error if things are commented out, but we can leave the keyvault files as is. Does anyone have a preference on how to unlink some code for the moment until we need keyvault again? cc @sarath-p @evanlouie @bnookala

I will remove the code. I also think that it is right thing to do

andrebriggs (approved):

src/commands/deployment/onboard.decorator.json
samiyaakhtar: The presence of these in config file makes them look mandatory for other commands as well. Perhaps in future we should revisit whether or not the config file should have four of these.

samiyaakhtar (approved):

src/commands/deployment/onboard.decorator.json
dennisseah: these are not in config file, these are option values for onboard command only

samiyaakhtar: they're also in config file: https://github.com/CatalystCode/spk/blob/master/spk-config.yaml#L32

dennisseah: i missed these, should I remove them to avoid confusion?

samiyaakhtar: I think for now we can leave them since I believe the onboard command takes them either through config file or through the command itself. Although since it's a one time command, it shouldn't rely on the config file. I'll open a separate discussion for that :)

dennisseah: ok

sarath-p: The service principal related options are used to authenticate with Azure to create storage account and table if they don't exist. Deleting these options make this command not consistent with other commands to pass all required values without depending on config.yaml.

PR #462, opened 2020-03-27
Adding design doc template

andrebriggs:


docs/commands/data.json
```diff
@@ -191,10 +191,6 @@
         "description": "Name of the resource group to create new storage account when it does not exist",
         "required": true
       },
-      {
```
NathanielRose: Was this intentionally made to be modified?

NathanielRose (approved):

docs/commands/data.json
andrebriggs: @NathanielRose we have a docs regeneration pre-commit hook that is updating the docs

dennisseah: @NathanielRose 
this is generated by pre-commits hooks. it is ok

dennisseah:
I think we can add Appendix section where we list down (if any) url to documents/articles/guides that aid in writing this doc. what do you think?

andrebriggs:
> I think we can add Appendix section where we list down (if any) url to documents/articles/guides that aid in writing this doc. what do you think?

@dennisseah Good idea. I'll update

samiyaakhtar (approved):

dennisseah (approved):

PR #463, opened 2020-03-27
Disable camelCase ESLint Rule

evanlouie:
Disabling the `@typescript-eslint/camelcase` ESLint rule because we store a lot
of config level objects use underscore case and we are most likely not going to
get away from that.

- Disabled the ESLint `@typescript-eslint/camelcase` rule in `.eslintrc.js`.
- Removed all page-level `@typescript-eslint/camelcase` disable flags.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,4 +12,7 @@ module.exports = {
     "prettier",
     "prettier/@typescript-eslint",
   ],
+  rules: {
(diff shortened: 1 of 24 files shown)


samiyaakhtar (approved):

dennisseah (approved):
thanks Evan

andrebriggs (approved):

PR #471, opened 2020-03-30
[FEATURE] Add Specific error code for infra generate command

### LINKED ISSUE #1277: Specific error codes for infra generator command
**As a:**
User

**I want:**
error code for infra generate command

**So that:**
I can understand the root cause.

add error code and maintain exception chain

**Acceptance Criteria:**

**Additional context:**

No

part of https://github.com/microsoft/bedrock/issues/1088

dennisseah:
closes https://github.com/microsoft/bedrock/issues/1277

getting messages like this

```
$ ts-node src/index.ts infra generate 
error:   
code: 1000
message: infra-generate-cmd-failed: Infra generate Command was not successfully executed.
  code: 1010
  message: infra-defn-yaml-not-found: definition.yaml was not found in /Users/veseah/git/spk
```


dennisseah:
@yradsmikham, @NathanielRose there are two `TOFIX` in the changes that I cannot figure out, can you pls help? 
```
// TOFIX: are these warnings? if yes, please write some comments.
```

jmspring (approved):
Lgtm

src/commands/infra/generate.ts
```diff
@@ -353,13 +360,15 @@ export const generateConfigWithParentEqProjectPath = async (
     writeTfvarsFile(spkTfvarsObject, parentDirectory, SPK_TFVARS);
     await copyTfTemplate(templatePath, parentDirectory, true);
   } else {
+    // TOFIX: are these warnings? if yes, please write some comments.
     logger.warn(`Variables are not defined in the definition.yaml`);
   }
```
yradsmikham: Yes, this is a warning.

src/commands/infra/generate.ts
```diff
@@ -353,13 +360,15 @@ export const generateConfigWithParentEqProjectPath = async (
...
   if (parentInfraConfig.backend) {
     const backendTfvarsObject = generateTfvars(parentInfraConfig.backend);
     checkTfvars(parentDirectory, BACKEND_TFVARS);
     writeTfvarsFile(backendTfvarsObject, parentDirectory, BACKEND_TFVARS);
   } else {
+    // TOFIX: are these warnings? if yes, please write some comments.
```
yradsmikham: This is also another valid warning.

yradsmikham: Not all templates will require a remote backend (i.e Bedrock's azure-simple). Keep in mind, that if a remote backend is not configured for a template, it will be impossible to be able to use `spk infra` in a pipeline, but this can still work locally.

src/commands/infra/generate.ts
yradsmikham: Consider the case where the only common configuration is just backend configuration, and no common variable configuration. Thus, it is not "necessary" for a parent `definition.yaml` to have a variables block in a multi-cluster.

@NathanielRose do you agree?

NathanielRose: Agreed, we should be able to support common backend configs without enforcing common config between leaf and parent.

NathanielRose (approved):
LGTM

yradsmikham (approved):

dennisseah:
@yradsmikham and @NathanielRose , thank so much

PR #479, opened 2020-03-31
[HOUSEKEEPING] remove page wide eslint disable in project commands

dennisseah:


andrebriggs (approved):

andrebriggs (approved):

mtarng (approved):
Also looks like we're creating an interface ConfigValue to use in project/pipeline/ts

PR #480, opened 2020-03-31
Design doc for `spk service create`

yradsmikham:
Related to https://github.com/microsoft/bedrock/issues/1006.


edaena (approved):

jmspring (approved):
lgtm - my questions were answered in the design review

samiyaakhtar (approved):

technical-docs/designs/services/acrBuildArguments.md
```diff
@@ -0,0 +1,91 @@
...
+
+```sh
+spk service create "fabrikam.acme.frontend" -d "services" -p "chart" -g "https://dev.azure.com/bedrock/fabrikam/_git/fabrikam2019" -b "master" --build-vg "my-azdo-build-vg"
+```
+
+- Add argument `--build-variables` to take in an array of key-value pairs for
+  all variables to be used in a new variable group. An example would be
+
+```sh
+spk service create "fabrikam.acme.frontend" -d "services" -p "chart" -g "https://dev.azure.com/bedrock/fabrikam/_git/fabrikam2019" -b "master" --build-variables [version:0.0.1,path:$HOME]
```
sarath-p: The `key:value` pair supports non secret type variable by default. Does it need to support secret variable type? If so, it needs to be extended `key:value:type`  or `key:value:isSecret`.

mtarng (commented):
So the new flags `--build-vg` and `--build-variables` should be used in conjunction, and we won't be piggy-backing off another vg elsewhere.

Would we add these VG to the azdo project through `spk project create-variable-group` or `spk variable-group create`?

Additionally, I think we may need to expand the section about implementation:
>we can add logic to iterate through the a
linked variable group and inject variables as arguments to be part of the
`az acr build` command

We will also need to expose each variable group variable as an environment variable in that pipeline step.

technical-docs/designs/services/acrBuildArguments.md
yradsmikham: That's a good point. I will update the docs.

yradsmikham:
> So the new flags `--build-vg` and `--build-variables` should be used in conjunction, and we won't be piggy-backing off another vg elsewhere.
> 
> Would we add these VG to the azdo project through `spk project create-variable-group` or `spk variable-group create`?
> 
> Additionally, I think we may need to expand the section about implementation:
> 
> > we can add logic to iterate through the a
> > linked variable group and inject variables as arguments to be part of the
> > `az acr build` command
> 
> We will also need to expose each variable group variable as an environment variable in that pipeline step.

Is there a preference as to which spk `variable-group` command we should be using for this situation?

andrebriggs (commented):
> As part of the build-update-hld.yaml, where there is the "ACR Build and Publish" task (L208-L217) generated by spk service create, we can add logic to iterate through the a linked variable group and inject variables as arguments to be part of the az acr build command (i.e.az acr build -r $(ACR_NAME) -f $(DOCKER_BUILD_FILE) --build-arg $key=${value}).

@yradsmikham When exactly do we look at the variables in the variable group? Seems like this would have to happen "client side" before we `git commit + push` the `build-update-hld.yaml` file?

If a users adds additional args to their variable group do they have to call `spk service create` again locally, the commit and push?

yradsmikham:
> > As part of the build-update-hld.yaml, where there is the "ACR Build and Publish" task (L208-L217) generated by spk service create, we can add logic to iterate through the a linked variable group and inject variables as arguments to be part of the az acr build command (i.e.az acr build -r $(ACR_NAME) -f $(DOCKER_BUILD_FILE) --build-arg $key=${value}).
> 
> @yradsmikham When exactly do we look at the variables in the variable group? Seems like this would have to happen "client side" before we `git commit + push` the `build-update-hld.yaml` file?
> 
> If a users adds additional args to their variable group do they have to call `spk service create` again locally, the commit and push?

@andrebriggs , we access the variables in the variable group when the services build pipeline (i.e. fabrikam.acme.frontend-pipeline) runs, or when the `build-update-hld.yaml` is executed. This should happen as we are running `az acr build` shown [here](https://github.com/CatalystCode/spk/blob/39286c2ed3508588cb6aa2d1c249855dc3f4e1af/src/lib/fileutils.ts#L224).

If users make updates to their Dockerfile and require to add additional args to their variable group, they should (for now) just update their `build-update-hld.yaml` manually. This would mean adding a line to `export` the variable in the script, and adding another argument `--build-arg $key=$value`. 

Thoughts? @jmspring @mtarng

yradsmikham:
> So the new flags `--build-vg` and `--build-variables` should be used in conjunction, and we won't be piggy-backing off another vg elsewhere.
> 
> Would we add these VG to the azdo project through `spk project create-variable-group` or `spk variable-group create`?
> 
> Additionally, I think we may need to expand the section about implementation:
> 
> > we can add logic to iterate through the a
> > linked variable group and inject variables as arguments to be part of the
> > `az acr build` command
> 
> We will also need to expose each variable group variable as an environment variable in that pipeline step.

Sure, I can be a little more specific about this.

mtarng (approved):
Looks good. Thanks @yradsmikham

mtarng:
> > > As part of the build-update-hld.yaml, where there is the "ACR Build and Publish" task (L208-L217) generated by spk service create, we can add logic to iterate through the a linked variable group and inject variables as arguments to be part of the az acr build command (i.e.az acr build -r $(ACR_NAME) -f $(DOCKER_BUILD_FILE) --build-arg $key=${value}).
> > 
> > 
> > @yradsmikham When exactly do we look at the variables in the variable group? Seems like this would have to happen "client side" before we `git commit + push` the `build-update-hld.yaml` file?
> > If a users adds additional args to their variable group do they have to call `spk service create` again locally, the commit and push?
> 
> @andrebriggs , we access the variables in the variable group when the services build pipeline (i.e. fabrikam.acme.frontend-pipeline) runs, or when the `build-update-hld.yaml` is executed. This should happen as we are running `az acr build` shown [here](https://github.com/CatalystCode/spk/blob/39286c2ed3508588cb6aa2d1c249855dc3f4e1af/src/lib/fileutils.ts#L224).
> 
> If users make updates to their Dockerfile and require to add additional args to their variable group, they should (for now) just update their `build-update-hld.yaml` manually. This would mean adding a line to `export` the variable in the script, and adding another argument `--build-arg $key=$value`.
> 
> Thoughts? @jmspring @mtarng

Though according to docs [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/library/variable-groups?view=azure-devops&tabs=yaml#use-a-variable-group), once we have a reference to the variables, we could use a script to iterate through them, making secret substitutions and passing them all into the build args. So as long as the variable group variable is updated somehow (either though a future spk command or the azdo UI) then the build script should have the latest arguments.

PR #486, opened 2020-03-31
[HOUSEKEEPING] remove eslint disable from deployment get command

dennisseah:
reordering functions in this file to satisfy no-use-before-define eslint rule. no changes in logics/flow


src/commands/deployment/get.ts
```diff
@@ -296,74 +142,70 @@ export const getClusterSyncStatuses = (
...
-    !config.introspection.azure.table_name ||
-    !config.introspection.azure.key ||
-    !config.introspection.azure.partition_key ||
-    !config.introspection.azure.key
-  ) {
-    throw Error(
-      "You need to run `spk init` and `spk deployment onboard` to configure `spk."
-    );
+    if (repo) {
+      const promise = fetchPR(repo, strPr, initObj.accountName);
```
samiyaakhtar: initObj.accessToken

samiyaakhtar (approved):

src/commands/deployment/get.ts
dennisseah: accept. Do not know why git does not bring over your changes. Maybe the code got moved around too much

andrebriggs (approved):

PR #493, opened 2020-04-01
resolve optional arguments for spk pipeline commands

### LINKED ISSUE #1279: Spk commands to install pipelines should not be enforcing requirements for optional parameters
**Describe the bug:**
spk pipeline stall commands like `spk project install-lifecycle-pipeline`, `spk service install-build-pipeline` and maybe: `spk hld install-manifest-pipeline` are enforcing checks in the command decorator `execute` function.

Repository Url and pipeline names are optional parameters, they should not be required.

**To Reproduce:**
- Initialize spk with a config file
- Create and initialize a project via `spk project init` and `spk project create-variable-group`
- Try to install the project lifecycle pipeline without any flags: `spk project install-lifecycle-pipeline`. It will error because it expects a repo url and pipeline name to be passed.

**Expected behavior:**
Pipeline commands should be able to run without flags if a config file is configured for spk.

No

yradsmikham:
Close https://github.com/microsoft/bedrock/issues/1279.


src/commands/project/pipeline.ts
```diff
@@ -66,6 +66,26 @@ export const checkDependencies = (projectPath: string): void => {
   }
 };
 
+export const validateRepoUrl= (
```
samiyaakhtar: nit: jsdoc

samiyaakhtar (approved):

src/commands/project/pipeline.ts
```diff
@@ -66,6 +66,26 @@ export const checkDependencies = (projectPath: string): void => {
...
+        return repoUrl = getRepositoryUrl(gitOriginUrl)
+      }
+    } else {
+      return repoUrl = opts.repoUrl
+    }
+  } catch(err){
+    throw new Error(`Unable to obtain and validate the repo url.`)
+  }
+}
+
```
bnookala: just a stylistic comment - in some places there's uses of `;` at the end of every statement (see line 253 below), but there there isn't  any uses cases in this code block

I think we should _probably_ enable this eslint rule https://eslint.org/docs/rules/semi to settle it one way or another, but would just like to call it out as a difference

sarath-p (approved):

src/commands/project/pipeline.ts
```diff
@@ -66,6 +66,26 @@ export const checkDependencies = (projectPath: string): void => {
   }
 };
 
+export const validateRepoUrl= (
+  opts: CommandOptions,
+  gitOriginUrl: string
+): string => {
+  try {
+    let repoUrl = ''
+    if (!opts.repoUrl) {
```
bnookala: i think this logic could be made simpler by inverting the repoUrl check condition and returning early and only calling `getRepositoryUrl` once:

```
if (opts.repoUrl) {
    return opts.repoUrl;
}

const repoUrl = getRepositoryUrl(gitOriginUrl);

if (!repoUrl) {
    throw Error(`Repo url not defined. Are you in a spk project folder?`);
}

return repoUrl;
```

bnookala (changes requested):

mtarng (changes requested):
We need to update the same logic for spk service install-build-pipeline too: https://github.com/CatalystCode/spk/blob/master/src/commands/service/pipeline.ts#L77-L79

src/commands/project/pipeline.ts
yradsmikham: I'm not sure if what you're proposing is the same logic I am trying to have the code achieve. What I am trying to do is check for the the argument value that is passed in for `repoUrl`, if it exist then set variable and proceed, skipping the need to confirm with `gitOriginUrl`. If it doesn't exist then check if it is possible obtain the `repoUrl` using `gitOriginUrl`. If that is not possible then throw an error, and if possible then set the variable appropriately. 

From your code, it seems like we first check to see if the argument was passed in, then we set the variable using `gitOriginUrl`. But we never prioritize the passed in argument to set the variable, and always rely on `gitOriginUrl` to determine the repo url, which I believe is not what we want.

src/commands/project/pipeline.ts
yradsmikham: Will enable eslint rule for consistency! Thanks!

yradsmikham:
> We need to update the same logic for spk service install-build-pipeline too: https://github.com/CatalystCode/spk/blob/master/src/commands/service/pipeline.ts#L77-L79

Resolved.

src/commands/service/pipeline.ts
```diff
@@ -33,6 +33,9 @@ import {
   validateOrgNameThrowable,
   validateProjectNameThrowable,
 } from "../../lib/validator";
+import { validateRepoUrl } from "../project/pipeline";
```
mtarng: If we're sharing this method, we should put it in one of our shared util files instead.

src/commands/project/pipeline.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
+/*eslint semi: "error"*/
```
dennisseah: we have attempting to remove page wide eslint-disable, please do not add them. Please contact @evanlouie or/and @bnookala if we want to semi: error in .eslintrc thanks

src/commands/project/pipeline.ts
yradsmikham: @bnookala I tested your snippet of code, and it appears to work! Thanks for the feedback. Marking resolved.

src/commands/project/pipeline.ts
yradsmikham: Will remove!

src/commands/project/pipeline.ts
```diff
@@ -97,13 +133,7 @@ export const fetchValidateValues = (
     );
   }
   const azureDevops = spkConfig?.azure_devops;
-  if (!opts.repoUrl) {
-    throw buildError(
-      errorStatusCode.VALIDATION_ERR,
-      "project-pipeline-err-repo-url-undefined"
-    );
-  }
```
dennisseah: can you help to remove `project-pipeline-err-repo-url-undefined` from src/lib/i18n.json? thanks

src/commands/project/pipeline.ts
```diff
@@ -76,6 +77,41 @@ export const checkDependencies = (projectPath: string): void => {
...
+  opts: CommandOptions,
+  gitOriginUrl: string
+): string => {
+  try {
+    if (opts.repoUrl) {
+      return opts.repoUrl;
+    }
+    const repoUrl = getRepositoryUrl(gitOriginUrl);
+    if (!repoUrl) {
+        throw Error(`Repo url not defined. Are you in a spk project folder?`);
```
dennisseah: this error is lost when we throw an exception here and then catch it later (at line 111) where it throws a new error.

dennisseah (commented):
a few comments, the rest looks good. thanks

src/commands/project/pipeline.ts
yradsmikham: is there a reason why we are removing it?

yradsmikham: Got it, will do.

src/commands/service/pipeline.ts
yradsmikham: I've added this to the gitutils.ts

dennisseah (approved):

src/lib/gitutils.ts
```diff
@@ -345,6 +346,27 @@ export const checkoutCommitPushCreatePRLink = async (
...
 };
 
+/**
+ * Returns a git repository url
+ *
+ * @param opts Options object from commander.
+ * @param gitOriginUrl Git origin URL which is used to set values
+ *        for pipeline, repoName and repoUrl
+ */
+export const validateRepoUrl = (
```
mtarng: Looks good. Just need a unit test for this method

yradsmikham: Done!

mtarng (approved):

bnookala (approved):

src/commands/project/pipeline.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
evanlouie: Why was this added back? Can't see any null assertions in the PR

yradsmikham: Hm, not sure. I removed it again. Something probably happened during rebase. Good eye!

PR #494, opened 2020-04-01
Add error codes in reconcile

edaena:
https://github.com/microsoft/bedrock/issues/1293


src/commands/hld/reconcile.ts
```diff
@@ -70,8 +72,16 @@ export const execAndLog = async (commandToRun: string): Promise<ExecResult> => {
...
-    throw result.error;
+    const errorInfo = buildError(
+      errorStatusCode.CMD_EXE_ERR,
+      {
+        errorKey: "hld-reconcile-err-cmd-failed",
+        values: [commandToRun],
+      },
+      result.error
+    );
+    logError(errorInfo);
```
dennisseah: i think we do not need to logError because this error will appear at the end as a exception chain

src/commands/hld/reconcile.ts
```diff
@@ -107,10 +117,16 @@ export const createRepositoryComponent = async (
...
+    const errorInfo = buildError(
+      errorStatusCode.EXE_FLOW_ERR,
+      {
+        errorKey: "hld-reconcile-err-repo-create",
+        values: [repositoryName, absHldPath],
+      },
+      err
     );
-    throw err;
+    logError(errorInfo);
```
dennisseah: same here

mtarng (approved):

bnookala (approved):

dennisseah (approved):
LGTM

PR #496, opened 2020-04-02
Fix repository name not populated

edaena:
Repository name is not part of the supported "opts"

### DIFF
--- a/src/commands/project/pipeline.ts
+++ b/src/commands/project/pipeline.ts
@@ -140,14 +140,14 @@ export const fetchValidateValues = (
   // values are validate, adding || "" is just to
   // satisfy the no-non-null-assertion eslint rule
   const configVals: ConfigValues = {
-    orgName: opts.orgName || "",
-    buildScriptUrl: opts.buildScriptUrl || BUILD_SCRIPT_URL,
-    devopsProject: opts.devopsProject || "",
+    orgName: values.orgName || "",
+    buildScriptUrl: values.buildScriptUrl || BUILD_SCRIPT_URL,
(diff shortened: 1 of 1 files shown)


samiyaakhtar (approved):

src/commands/project/pipeline.ts
```diff
@@ -143,7 +143,7 @@ export const fetchValidateValues = (
     orgName: opts.orgName || "",
```
mtarng: This whole logic section is kinda confusing. Why are we still loading from `opts` from here. Shouldn't all of these be loaded from the `values` object instead? 

This was introduced in: https://github.com/CatalystCode/spk/pull/479/files

@edaena @dennisseah does  this make sense for us to fix this all here? Does similar logic for other ConfigValues interfaces exist now too in other commands?

mtarng (changes requested):

src/commands/project/pipeline.ts
dennisseah: @mtarng do you know other places that this problem too?

dennisseah (approved):

src/commands/project/pipeline.ts
mtarng: I'm not sure, which is why I asked the question @dennisseah . Have we refactored any other commands to move arguments from `CommandOptions` to `ConfigValues`?

mtarng (approved):

src/commands/project/pipeline.ts
dennisseah: I will double check

PR #499, opened 2020-04-02
bedrock.yaml refactor & service create spec update

### LINKED ISSUE #1117: Update `service create` parameters and update bedrock.yaml schema
This is the first task for refactoring how we handle service references in
bedrock.yaml

**I want:** `spk service create` and `bedrock.yaml` to be reworked to index
based on a serviceName

**So that:**

There is less confusion about a service's name vs a service's path

- Rewrite `spk service create` to accept to positional arguments
  `spk service create <svc-name> <path-to-svc>`
- Change the schema for `bedrock.yaml` services to index on name and have a
  mandatory `path` parameter.
- Automatically migrate the old bedrock.yaml services schema to the new one
  whenever the an `spk service` command is run
  - Use the new required parameter `path` in the service definition as the
    differentiator to detect when the service is the old schema or the new
    schema.

**Acceptance Criteria:**

Refer to above section.

**Additional context:**

Parent epic: https://github.com/microsoft/bedrock/issues/943

Yes -- update the docs to reflect the new `spk service create` command arguments

evanlouie:
- Updated bedrock.yaml `services` to contain a list of services instead of a map
  indexed by path.
  - Auto-migration of bedrock.yaml is hooked into `read()` function of the
    `bedrockYaml.ts` file.
    - `read()` is now the only function used for loading the bedrock.yaml file;
      `Bedrock()` in `config.ts` is now noted as deprecated and now calls
      `read()` under the hood.
  - `path` property for a service now tracks the path to the service.
  - Documentation references all updated.
- `spk service create <service-name>` now follows the spec
  `spk service create <service-name> <service-path>`
  - Documentation references all updated.
  - Integration tests updated.

closes https://github.com/microsoft/bedrock/issues/1117


mtarng:
We'll also have to follow up with documentation for bedrock once we cut a release. We'll probably have to go to version 0.6.0 or 0.7.0 depending if 0.6.0 is for rings or not.

evanlouie:
> We'll also have to follow up with documentation for bedrock once we cut a release. We'll probably have to go to version 0.6.0 or 0.7.0 depending if 0.6.0 is for rings or not.

You mean documentation regarding the changes? I've updated the docs to reflect the spec change

mtarng:
> > We'll also have to follow up with documentation for bedrock once we cut a release. We'll probably have to go to version 0.6.0 or 0.7.0 depending if 0.6.0 is for rings or not.
> 
> You mean documentation regarding the changes? I've updated the docs to reflect the spec change

Right, I mean the walkthrough documentation that's in the bedrock repository.

samiyaakhtar (approved):
LGTM - Tested this for the most part while working on https://github.com/microsoft/bedrock/issues/1173

yradsmikham (approved):

PR #505, opened 2020-04-03
hld lifecycle pipeline to blow away repository directory in hld before regenerating assets

### LINKED ISSUE #1118: spk reconcile should remove rings from hld once they've been remove from a project
**As a:**
spk user
**I want:**
my hld repository to be reconciled once i remove a ring from my project

when i remove a ring via `spk ring delete`, I expect the project lifecycle pipeline to:
- [ ] For each linked service in this project's `bedrock.yaml`, in the HLD repository:
  - [ ] Remove the ring from the service's `component.yaml`
  - [ ] Remove the ring directory from the service's directory.

**Acceptance Criteria:**
See above tasks.

Integration test should be handled by:
https://github.com/microsoft/bedrock/issues/974

**Additional context:**
ring management epic: https://github.com/microsoft/bedrock/issues/955

yes!

mtarng:
hld lifecycle pipeline will remove all hld components for a repository except files in /config folders and `access.yaml` files.

This allows the lifecycle pipeline to remove rings and possibly services as well.

Closes: https://github.com/microsoft/bedrock/issues/1118


src/commands/hld/reconcile.test.ts
```diff
@@ -169,6 +172,109 @@ describe("createAccessYaml", () => {
...
+      expect.any(Function)
+    );
+  });
+
+  it("should throw an error if fs fails", async () => {
+    fsSpy.mockImplementationOnce(() => {
+      throw Error("some error");
+    });
+
+    expect(() => purgeRepositoryComponents(hldPath, repositoryName)).toThrow;
```
dennisseah: toThrow -> toThrow()

src/lib/ioUtil.ts
```diff
@@ -68,3 +69,27 @@ export const getMissingFilenames = (
     )
     .map((f) => path.basename(f));
 };
+
+/**
+ * Get all files/filepaths in this directory and subdirectories.
```
dennisseah: nit: Returns all the files/filepaths ...

src/commands/hld/reconcile.ts
```diff
@@ -101,6 +102,39 @@ type MiddlewareMap<T = Partial<ReturnType<typeof middleware.create>>> = {
...
+
+  try {
+    filesToDelete.forEach((file) => {
+      fs.unlink(file, function (err) {
+        if (err) throw err;
+        console.log(`${file} deleted!`);
+      });
+    });
+  } catch (e) {
+    logger.error(
```
edaena: Use buildError to create the error and logError to log it. Throw the buildError.

src/commands/hld/reconcile.ts
```diff
@@ -101,6 +102,39 @@ type MiddlewareMap<T = Partial<ReturnType<typeof middleware.create>>> = {
   default?: T;
 };
 
+// In spk hld reconcile, the results should always result in the same artifacts being created based on the state of bedrock.yaml.
+// The only exception is for files under the /config directories and any access.yaml files.
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
edaena: Description of parameters is missing

src/commands/hld/reconcile.ts
```diff
@@ -101,6 +102,39 @@ type MiddlewareMap<T = Partial<ReturnType<typeof middleware.create>>> = {
...
+  repositoryName: string
+): void => {
+  assertIsStringWithContent(absHldPath, "hld-path");
+  assertIsStringWithContent(repositoryName, "repository-name");
+
+  const filesToDelete = getAllFilesInDirectory(
+    path.join(absHldPath, repositoryName)
+  ).filter(
+    (filePath) =>
+      !filePath.match(/access\.yaml$/) && !filePath.match(/config\/.*\.yaml$/)
```
dennisseah: just a note. regular expression uses more resource (CPU), we try to avoid it if we can. e.g. `filePath.match(/access\.yaml$/)` can be `file.endsWith(access.yaml)

src/commands/hld/reconcile.ts
dennisseah: replace 
```
logger.error(`error purging repository component '${repositoryName}' in path '${absHldPath}'`);
throw e;
```
with
```
      throw buildError(
        errorStatusCode.FILE_IO_ERR,
        {
          errorKey: "hld-reconcile-err-purge-repo-comps",
          values: [repositoryName, absHldPath],
        },
        err
      );
```
and add to `lib/i18n.json`

```
   "hld-reconcile-err-purge-repo-comps": "Could not purge repository component {0} in path {1}."
```

when the error happen, we will get
```
code: 1011
message: hld-reconcile-err-purge-repo-comps: Could not purge repository component <something> in path <some path>.
details: <error message from err>
```

We do not need to do `logger.err` anymore. The entire chain of error will appear at the end of spk.log and console. So we can see the execution path and where error is coming from

edaena: You are correct, we can just throw the buildError, no need to log it in this case.

src/commands/hld/reconcile.ts
```diff
@@ -101,6 +102,39 @@ type MiddlewareMap<T = Partial<ReturnType<typeof middleware.create>>> = {
...
+  absHldPath: string,
+  repositoryName: string
+): void => {
+  assertIsStringWithContent(absHldPath, "hld-path");
+  assertIsStringWithContent(repositoryName, "repository-name");
+
+  const filesToDelete = getAllFilesInDirectory(
+    path.join(absHldPath, repositoryName)
+  ).filter(
+    (filePath) =>
```
bnookala, line 121: wondering if there's any other places where we _shouldnt_ delete as well ie static?

mtarng: Hm, for now we're generating all static components every time. What are manual configurations that would be made in static?

bnookala: thinking that if we've advised users to add manifests in static to be applied against a cluster, it might be a breaking change

dennisseah (approved):

samiyaakhtar (approved):

PR #513, opened 2020-04-06
Add markup file for hld command

edaena:


src/commands/hld/append-variable-group.md
```diff
@@ -0,0 +1,43 @@
...
+pool:
+  vmImage: ubuntu-latest
+steps:
+.
+.
+.
+```
+
+running `spk hld append-variable-group my-vg` with a variable group name, in
+this case `my-vg`, will add it under the `variables` section:
```
dennisseah: what happen when variable groups in the yaml already exist? Do we add or we have logic to skip?

samiyaakhtar (approved):

src/commands/hld/append-variable-group.md
edaena: No, there's no logic to verify, it will add a duplicate. The pipeline will not be affected. It will still succeed.

sarath-p (approved):

src/commands/hld/append-variable-group.md
andrebriggs: @edaena let's add logic to prevent that. Also we probably should mention this behavior in the MD file

yradsmikham (approved):

PR #518, opened 2020-04-07
Adding spk service get-display-name command

samiyaakhtar:
This is part 1 of https://github.com/microsoft/bedrock/issues/1173. The second part will be running this command as part of pipelines to get the right service name, but that will be merged after an spk release is available with the changes in this PR. 

Screen Shot 2020-04-07 at 7 06 50 PM

Related to https://github.com/microsoft/bedrock/issues/1173


docs/commands/data.json
```diff
@@ -499,19 +505,19 @@
...
...     git: \"https://dev.azure.com/fabrikam/frontend/_git/charts\"\n        path: frontend\n    k8sBackend: \"fabrikam-k8s-svc\"\n    k8sBackendPort: 80\n    middlewares: []\n    pathPrefix: \"fabrikam-service\"\n    pathPrefixMajorVersion: \"v1\"\nvariableGroups:\n  - fabrikam-vg\n```\n\nrunning `spk ring set-default prod` will result in:\n\n1. `prod` will be set as the default in `bedrock.yaml`:\n\n   ```yaml\n   rings:\n   dev:\n   qa:\n   prod:\n     isDefault: true\n   services:\n     - path: ./\n       displayName: \"fabrikam\"\n       helm:\n       chart:\n         branch: master\n         git: \"https://dev.azure.com/fabrikam/frontend/_git/charts\"\n         path: frontend\n       k8sBackend: \"fabrikam-k8s-svc\"\n       k8sBackendPort: 80\n       middlewares: []\n       pathPrefix: \"fabrikam-service\"\n       pathPrefixMajorVersion: \"v1\"\n   variableGroups:\n     - fabrikam-vg\n   ```\n\n2. Commiting these changes will trigger the project's lifecycle pipeline, which\n   may update the default \"no-ring\" IngressRoutes to route to the `prod` service\n   in the linked HLD repository,\n   [pending the work in this issue](https://github.com/microsoft/bedrock/issues/1084).\n"
```
samiyaakhtar: trying to get rid of these changes from my branch but the pre-commit hook keeps adding it back

src/commands/project/get-display-name.ts
```diff
@@ -0,0 +1,62 @@
...
+    const bedrockFile = readBedrockYaml(currentDirectory);
+    const servicePath = process.cwd().replace(currentDirectory, ".");
+    for (const serviceIndex in bedrockFile.services) {
+      if (servicePath === bedrockFile.services[serviceIndex].path) {
+        console.log(bedrockFile.services[serviceIndex].displayName);
+        await exitFn(0);
+      }
+    }
+    // Ideally, we should not be getting here, so throw an error.
+    throw new Error("Display name for current directory could not be found.");
```
mtarng: don't need `new`. Also we're moving our thrown errors into a new pattern. An example: https://github.com/CatalystCode/spk/pull/505#pullrequestreview-387384084

src/commands/project/get-display-name.ts
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable @typescript-eslint/no-use-before-define */
```
mtarng: we can probably remove this and fix the eslint rule `no-use-before-define`

src/commands/project/get-display-name.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @param opts validated option values
+ * @param exitFn exit function
+ */
+export const execute = async (
+  opts: CommandOptions,
+  exitFn: (status: number) => Promise<void>
+): Promise<void> => {
+  let currentDirectory = process.cwd() + "/bedrock.yaml";
+  try {
+    // First, find a bedrock.yaml file in this directory and if not, iterate by
```
mtarng: Hm. Let's not do this tree climbing. We're pretty strict about running `spk project` and `spk service` commands only in the root of the project repository.

mtarng (changes requested):
Hm. Overall I think this logic should probably fall under `spk service` commands, and there should be a parameter for the service "index" or path. 

Is this command designed to run in the service build pipeline, and so the design assumes you're in the service directory?

src/commands/project/get-display-name.ts
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable @typescript-eslint/no-use-before-define */
+/* eslint-disable @typescript-eslint/camelcase */
```
dennisseah: we do not need this. @evanlouie has disabled this globally

src/commands/project/get-display-name.ts
```diff
@@ -0,0 +1,62 @@
...
+    // Ideally, we should not be getting here, so throw an error.
+    throw new Error("Display name for current directory could not be found.");
+  } catch (err) {
+    logger.error(err);
+    await exitFn(1);
+  }
+};
+
+/**
+ * Adds the validate command to the commander command object
```
dennisseah: nit; doc is incorrect

samiyaakhtar:
> Is this command designed to run in the service build pipeline, and so the design assumes you're in the service directory?
@mtarng yes, it's gonna be run from build-update-hld.yaml so it will be in the service directory, it looks for a bedrock.yaml in directories above it and extracts the service name from it. Should it be moved under `spk service` instead? Now that I think about it, i think it should be moved there 🤔

src/commands/project/get-display-name.ts
samiyaakhtar: @mtarng this command is not meant so much for users to run manually, but just for the pipeline itself, to extract the name and send to introspection storage.  I'm thinking if it should even be baked into spk and available for the user to run as a command or find it confusing in docs on why it's even available, since it's so trivial. @andrebriggs thoughts?

docs/commands/data.json
dennisseah: you should not touch this file. the content of this file is generated

samiyaakhtar: I didn’t, it automatically added these changes to this file

src/commands/project/get-display-name.ts
```diff
@@ -0,0 +1,62 @@
...
+      const split = currentDirectory.split("/");
+      if (split.length < 2) {
+        break;
+      }
+      currentDirectory = currentDirectory.replace(
+        split[split.length - 2] + "/",
+        ""
+      );
+    }
+    currentDirectory = currentDirectory.replace("/bedrock.yaml", "");
```
dennisseah: there are a few about this implementation. 
1. you add `bedrock.yaml` to the current directory and then you slice and dice to find the directory that contains `bedrock.yaml`. 
2. then you remove `bedrock.yaml` from the current directory :-)
3. as you are climbing the directory, you can likely to hit to a point where you get file permission error unless you are the root user. How are we going to handle this?
4. `currentDirectory = currentDirectory.replace("/bedrock.yaml", "");` you are making assumption that you found the directory where `bedrock.yaml` resides. What happen if this is not the case. then `readBedrockYaml(currentDirectory);` will fail. and user has no clues what is wrong.
5. you only have one test, can we have more tests?

i  think the code can be better written this way. Sorry I maybe wrong
```
export const locateFolder = (dir: string, fileName: string): string => {
  try {
    if (fs.existsSync(path.join(dir, fileName))) {
      return dir;
    }
  } catch (err) {
    throw buildError(<some message> // likely file permission error
  }

  const paths = currentDirectory.split(path.sep);
  if (paths.length < 2) {
    throw buildError(<some messages>
  }
  paths.pop();
  return locateFolder(paths.join(path.sep), fileName);
};

export const execute = async (
  opts: CommandOptions,
  exitFn: (status: number) => Promise<void>
): Promise<void> => {
  try {
    const bedrockYamlDir = locateFolder(process.cwd(), "bedrock.yaml");

    // now we are sure that bedrockYamlDir is good :-)
    // otherwise locateFolder would throw meaningful exception
    const bedrockFile = readBedrockYaml(bedrockYamlDir);

    const servicePath = process.cwd().replace(bedrockYamlDir, ".");
    
    // || {}, just in case bedrockFile.services is not defined
    // we do not need this if we are sure that bedrockFile.services
    // is always defined.
    const found = Object.keys(bedrockFile.services || {}).find((key) => {
      return servicePath === bedrockFile.services[key].path;
    });

    if (found) {
      console.log(bedrockFile.services(found).displayName);
      await exitFn(0);
    } else {
      throw buildError(...
    }
  } catch (err) {
    logError(...)
    await exitFn(1);
  }
};

```
write tests extensively for these 2 functions.

src/commands/project/get-display-name.ts
```diff
@@ -0,0 +1,62 @@
...
+    for (const serviceIndex in bedrockFile.services) {
+      if (servicePath === bedrockFile.services[serviceIndex].path) {
+        console.log(bedrockFile.services[serviceIndex].displayName);
+        await exitFn(0);
+      }
+    }
+    // Ideally, we should not be getting here, so throw an error.
+    throw new Error("Display name for current directory could not be found.");
+  } catch (err) {
+    logger.error(err);
```
dennisseah: log the entire error chain. example https://github.com/CatalystCode/spk/blob/master/src/commands/deployment/create.ts#L170

dennisseah (changes requested):

src/commands/project/get-display-name.ts
samiyaakhtar: thanks @dennisseah, I ended up changing the implementation quite a bit because the command should be ran from the root folder, and not service folder, that simplified the code. Please take a look again.

src/commands/service/get-display-name.ts
```diff
@@ -0,0 +1,72 @@
...
+ */
+export const execute = async (
+  opts: CommandOptions,
+  exitFn: (status: number) => Promise<void>
+): Promise<void> => {
+  // The assumption is that this command should be ran from the directory where bedrock.yaml exists
+  try {
+    if (!opts.path) {
+      throw buildError(
+        errorStatusCode.CMD_EXE_ERR,
```
dennisseah: `errorStatusCode.VALIDATION_ERR`, please do not have any code as CMD_EXE_ERR. please get familiarize with https://github.com/CatalystCode/spk/blob/master/src/lib/errorStatusCode.ts

src/commands/service/get-display-name.ts
```diff
@@ -0,0 +1,72 @@
...
+        "service-get-display-name-path-missing-param-err"
+      );
+    }
+    const bedrockFile = readBedrockYaml(process.cwd());
```
dennisseah, line 32: why `process.cwd()`? what happens to `opts.path`?

src/commands/service/get-display-name.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (!bedrockFile) {
+      throw buildError(
+        errorStatusCode.CMD_EXE_ERR,
```
dennisseah: `errorStatusCode.FILE_IO_ERR`

src/commands/service/get-display-name.ts
```diff
@@ -0,0 +1,72 @@
...
+      );
+    }
+    for (const serviceIndex in bedrockFile.services) {
+      if (opts.path === bedrockFile.services[serviceIndex].path) {
+        console.log(bedrockFile.services[serviceIndex].displayName);
+        await exitFn(0);
+      }
+    }
+
+    throw buildError(errorStatusCode.CMD_EXE_ERR, {
```
dennisseah: `errorStatusCode.ENV_SETTING_ERR`

dennisseah (changes requested):
where is `get-display-name.md`?
please write more tests, like
1. opts.path is incorrect.
2. yaml file does not have the service name.
please do not stop are only these two tests

thanks

src/commands/service/get-display-name.ts
mtarng: We're currently requiring that `service` and `project` commands be run only at the root of an application repo, so it's expected that there is a `bedrock.yaml` file in the current directory.

samiyaakhtar: @dennisseah `opts.path` is not for where bedrock.yaml is, it's used for finding a corresponding display name for a service that is at `opts.path`

dennisseah: got it. thanks

src/commands/service/get-display-name.md
```diff
@@ -0,0 +1,4 @@
+## Description
+
+Gets display name of a service based on the provided path by extracting this
+information from bedrrock.yaml
```
dennisseah: Gets display name of a service based on the provided path by extracting this
information from `bedrock.yaml`. This command tries to locate `bedrock.yaml` in the current folder, ...

please add more information like what happen of `bedrock.yaml` is not found, or have it as pre requisite that `bedrock.yaml` has to be in the current directory.

samiyaakhtar: @dennisseah Done. Please note that this command is not going to be useful for anyone to run manually (it's trivial), it's for the pipeline to run.

src/commands/service/get-display-name.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (!bedrockFile) {
+      throw buildError(
+        errorStatusCode.FILE_IO_ERR,
+        "service-get-display-name-bedrock-yaml-missing-err"
+      );
+    }
+    for (const serviceIndex in bedrockFile.services) {
+      if (opts.path === bedrockFile.services[serviceIndex].path) {
+        console.log(bedrockFile.services[serviceIndex].displayName);
+        await exitFn(0);
```
dennisseah: i am not comfortable with having command to finish in a for loop. please have it as (or other ways)
```
const found = Object.keys(bedrockFile.services).find(serviceIndex =>
      opts.path === bedrockFile.services[serviceIndex].path;

if (found) {
   console.log(bedrockFile.services[serviceIndex].displayName);
    await exitFn(0);
}
```

src/commands/service/get-display-name.ts
```diff
@@ -0,0 +1,75 @@
...
+        "service-get-display-name-bedrock-yaml-missing-err"
+      );
+    }
+
+    const serviceIndex = Object.keys(bedrockFile.services).find(
+      (index) => opts.path === bedrockFile.services[+index].path
+    );
+
+    if (serviceIndex) {
+      console.log(bedrockFile.services[+serviceIndex].displayName);
```
mtarng, line 45: Just had a though @samiyaakhtar , so `displayName` can be a an empty string. What do we want to return in this case? Is an empty string a valid case?

src/commands/service/get-display-name.test.ts
```diff
@@ -0,0 +1,34 @@
...
+beforeAll(() => {
+  enableVerboseLogging();
+});
+
+afterAll(() => {
+  disableVerboseLogging();
+});
+
+describe("get display name", () => {
+  it("positive test", async () => {
```
mtarng, line 17: Can we add negative tests? ie, no bedrock.yaml or no displayName found etc.

src/commands/service/get-display-name.ts
samiyaakhtar: @mtarng I think so, it should be valid but will just show an empty service name in the dashboard. I think that's ok since we want this to be the only source of truth for service name

mtarng: Sounds good.

evanlouie: Agree that a validation would be good. But should be noted that `spk service create` now requires a non-empty displayName to be provided

mtarng (approved):

src/commands/service/get-display-name.ts
samiyaakhtar: @evanlouie in that case we should be good going forward

yradsmikham (approved):

dennisseah (approved):

PR #527, opened 2020-04-07
Removing TODOs and simplifying deploymenttable.ts

### LINKED ISSUE #1301: Refactor deploymenttable.ts logic in CLI
Currently we have a handful of `TODO`s in this code base around how we link together pipelines. Consider another way to link together the pipelines or even how we do this telemetry altogether

[Here](https://github.com/CatalystCode/spk/blob/f1f90d02cb14bc34040e91a7c195ab5479214c72/src/lib/azure/deploymenttable.ts#L366)
```
      // 2. when cannot find the entry, we take the last row and INSERT it.
      // TODO: rethink this logic.
      return await updateLastRowOfArcToHLDPipelines(
```
[Here](https://github.com/CatalystCode/spk/blob/f1f90d02cb14bc34040e91a7c195ab5479214c72/src/lib/azure/deploymenttable.ts#L381)
```
    // Fallback: Ideally we should not be getting here, because there should
    // always be a p1 for any p2 being created.
    // TODO: rethink this logic.
    return await addNewRowToArcToHLDPipelines(
```

[Here](https://github.com/CatalystCode/spk/blob/f1f90d02cb14bc34040e91a7c195ab5479214c72/src/lib/azure/deploymenttable.ts#L657)
```
    // 2. when cannot find the entry, we take the last row and INSERT it.
    // TODO: rethink this logic.
    return await updateLastHLDtoManifestEntry(
```

[Here](https://github.com/CatalystCode/spk/blob/f1f90d02cb14bc34040e91a7c195ab5479214c72/src/lib/azure/deploymenttable.ts#L671)
```
  // Fallback: Ideally we should not be getting here, because there should
  // always matching entry
  // TODO: rethink this logic.
  return await addNewRowToHLDtoManifestPipeline(
```

samiyaakhtar:
Closes https://github.com/microsoft/bedrock/issues/1301


evanlouie (approved):

andrebriggs (changes requested):
Pending sync to discuss this area

yradsmikham (approved):

mtarng (commented):
Looks like we can probably also remove `/* eslint-disable @typescript-eslint/no-use-before-define */` from `deploymenttable.ts` at this time.

mtarng (approved):

dennisseah (approved):

andrebriggs (approved):

PR #528, opened 2020-04-08
[HOUSEKEEPING] have error identifiers for two errors in azure git lib

dennisseah:


yradsmikham (approved):

yradsmikham:
Is there is an issue associated with this?

dennisseah:
> Is there is an issue associated with this?

@yradsmikham, @andrebriggs and @evanlouie are ok with adding a card in https://github.com/microsoft/bedrock/projects/42 for small housekeeping tasks like this one.

src/lib/git/azure.ts
```diff
@@ -366,11 +366,10 @@ export const repositoryHasFile = async (
   );
 
   if (gitItem === null) {
-    throw Error(
-      "Error installing build pipeline. Repository does not have a " +
-        fileName +
-        " file."
-    );
+    throw buildError(errorStatusCode.GIT_OPS_ERR, {
+      errorKey: "git-azure-file-no-exist-in-repo",
```
bnookala, line 370: can we just call this git-azure-file-exist? i don't see a reason why we'd throw an error _if_ the file exists

src/lib/git/azure.ts
```diff
@@ -393,9 +392,10 @@ export const validateRepository = async (
   const repo = await gitApi.getRepository(repoName, project);
 
   if (!repo) {
-    throw Error(
-      `Project '${project}' does not contain repository '${repoName}'.`
-    );
+    throw buildError(errorStatusCode.GIT_OPS_ERR, {
+      errorKey: "git-azure-repo-no-exist",
```
bnookala, line 396: same feedback here

bnookala (changes requested):

src/lib/git/azure.ts
dennisseah:
```
  if (gitItem === null) {
    throw buildError(errorStatusCode.GIT_OPS_ERR, {
      errorKey: "git-azure-file-no-exist-in-repo",
      values: [fileName, branch, repoName],
    });
  }
```
if `gititem === null` means that file does not exist. Am I right? Or it is the other way around. `gititem === null` means file exist?

bnookala (approved):

PR #532, opened 2020-04-08
[INFRA] Support for local paths Nate.infra.relative paths

NathanielRose:
- Tasks of https://github.com/microsoft/bedrock/issues/1149
- Documentation for relative paths
- Implementation for local paths
- Unit Tests


src/commands/infra/generate.ts
```diff
@@ -501,6 +502,60 @@ export const singleDefinitionGeneration = async (
...
 
+/**
+ * Checks to see if module sources are local
+ *
+ * @param tfFile path to the terraform file in child directory
+ */
+export const checkModuleSource = (tfData: string): boolean => {
+  // Check if the file string matches an instance of a module source value as a local path
+  const matches = tfData.match(/^\s*source\s+=\s+["'](\.\.?\/[^"']*)["']$/gm);
+  if (matches != null) {
```
dennisseah: use `!==` and not `!=`

the code can be just
```
return matches !== null;
```

src/commands/infra/generate.ts
```diff
@@ -501,6 +502,60 @@ export const singleDefinitionGeneration = async (
...
+      const moduleSource = new RegExp(splitLine[3].replace(/['"]+/g, ""), "g");
+      // Get relative path of terraform module local to the repo
+      const repoModulePath = await simpleGit(
+        path.join(
+          sourcePath,
+          fileSource.template,
+          splitLine[3].replace(/["']/g, "")
+        )
+      ).revparse(["--show-prefix"]);
+      // Concatendate the Git URL with munged data
```
dennisseah: concatenate

src/commands/infra/generate.ts
```diff
@@ -576,6 +630,36 @@ export const generateConfig = async (
       templatePath
     );
   }
+  // TODO: Function to check for relative paths:
```
dennisseah: by adding this block of code into this `generateConfig` function shall make the function huge and you will find it hard to write unit test. I guess you move this block of code to a separate function

NathanielRose: Are you referring to lines 630 - 650? Im good with that -You're saying decouple from generateConfig and place the local path manipulation in a separate function.

dennisseah: yes: 630 - 654 to be precise. you can have these code in a new function and have `generateConfig` function to call this new function. The objective is to have smaller testable functions. thanks

src/commands/infra/generate.ts
```diff
@@ -23,6 +23,7 @@ import { copyTfTemplate } from "./scaffold";
 import { build as buildError, log as logError } from "../../lib/errorBuilder";
 import { errorStatusCode } from "../../lib/errorStatusCode";
 import { exec } from "../../lib/shell";
+import { file } from "mock-fs";
```
dennisseah: i guess this is added by accident

src/commands/infra/generate.ts
NathanielRose: Updated!

dennisseah (approved):

jmspring (approved):
Lgtm

src/commands/infra/generate.test.ts
```diff
@@ -778,3 +801,79 @@ describe("Validate backend.tfvars file", () => {
     );
   });
 });
+
+describe("test checkModuleSource function", () => {
+  it("positive test", () => {
+    let res = checkModuleSource(`source="../../azure/aks-gitops"`);
+    expect(res).toBeTruthy();
```
yradsmikham, line 808: Let's change this to `expect(res).toBe(true)` instead. This article is good at explaining why: https://vincenttunru.com/toBeTruthy-vs-toBe-true/

Same for the one below.

yradsmikham (changes requested):

yradsmikham:
Also, did we need to update the smoke tests?

src/commands/infra/generate.test.ts
dennisseah: checkModuleSource is returning boolean. using is toBeTruthy is ok.

NathanielRose:
> Also, did we need to update the smoke tests?

This is in a separate PR

andrebriggs (approved):

yradsmikham (approved):

PR #533, opened 2020-04-08
Add error codes to service commands

edaena:
Related to https://github.com/microsoft/bedrock/issues/1308


src/commands/service/create-revision.ts
```diff
@@ -69,12 +71,10 @@ export const getDefaultRings = (
...
   if (defaultRings.length === 0) {
-    throw Error(
-      `Default branches/rings must either be specified in ${join(
-        __dirname,
-        "bedrock.yaml"
-      )} or provided via --target-branch`
-    );
+    throw buildError(errorStatusCode.VALIDATION_ERR, {
+      errorKey: "service-create-revision-cmd-err-default-branch-missing",
+      values: [join(__dirname, "bedrock.yaml")],
```
dennisseah, line 76: i believe `join` should be `path.join`

src/lib/i18n.json
```diff
@@ -247,6 +247,11 @@
     "project-create-variable-group-cmd-err-values-missing": "Required values were missing. Provide values for registry-name, hld-repo-url, service-principal-id, service-principal-password, tenant.",
     "project-init-cmd-failed": "Project init was not successfully executed.",
 
+    "service-create-revision-cmd-failed": "Service create revision was not successfully executed.",
+    "service-create-revision-cmd-err-pr-source-branch": "A pull request for a branch cannot be made against itself. Ensure your target branch(es) '{0}' do not include your source branch '{1}'",
```
dennisseah, line 251: cannot -> could not

src/commands/service/create-revision.ts
```diff
@@ -167,17 +168,22 @@ export const execute = async (
...
 
     await makePullRequest(defaultRings, values);
     await exitFn(0);
   } catch (err) {
-    logger.error(err);
+    logError(
+      buildError(
+        errorStatusCode.CMD_EXE_ERR,
+        "service-create-revision-cmd-failed",
+        err
```
dennisseah, line 184: 👍

src/commands/service/create-revision.ts
bnookala: @dennisseah it's imported above from the module:

line 2:
```import { join } from "path";```

bnookala (approved):

src/commands/service/create-revision.ts
dennisseah: got it thanks

dennisseah (approved):

PR #534, opened 2020-04-08
[TEST] Added test for azure storage lib

dennisseah:
to bring the coverage to 90-ish%
also added error code for authentication with Azure DevOps with service principal


src/lib/azure/azurecredentials.ts
```diff
@@ -85,13 +87,21 @@ export const getManagementCredentials = async (
...
-    tenantId!
-  );
+  try {
+    // verifyConfigDefined has confirmed that these values are defined.
+    return await msRestNodeAuth.loginWithServicePrincipalSecret(
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      servicePrincipalId!,
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      servicePrincipalPassword!,
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
samiyaakhtar, line 97: Is there a better method to enforce non-null that does not require eslint-disable?

samiyaakhtar (approved):

src/lib/azure/azurecredentials.ts
dennisseah: i do not know. please let me know if you have some ideas

mtarng: @evanlouie thoughts?

mtarng (approved):

